PR #13, opened 2026-03-22
Document deletion, API routes, reasoning support

tobocop2:
### Document management
You can now delete documents from lilbee — not just add them. This works everywhere:
- `/delete` in the chat REPL with interactive file picker and tab completion
- REST API: `GET /api/documents` (paginated, searchable) + `POST /api/documents/remove`
- MCP: `lilbee_remove` and `lilbee_list_documents` tools

All deletion paths use a shared `remove_documents()` function with path traversal protection.

### Reasoning model support
Models like Qwen3 and DeepSeek-R1 wrap their thinking in `<think>...</think>` tags. lilbee now handles this:
- **Default (off):** Thinking is stripped — you see a clean answer
- **Enabled:** `LILBEE_SHOW_REASONING=true` — thinking appears in a muted style (CLI) or collapsible block (Obsidian)
- Settable via `/set show_reasoning true` in chat, env var, or config.toml

### API completeness
All server handlers now have HTTP routes:
- `GET /api/config` — all user-facing settings
- `GET /api/documents` — paginated document list with search
- Model catalog, install, show, delete endpoints

### Code quality
- Shared `remove_documents()` helper with path traversal guard
- Named types (`NodeSpan`, `RemoveResult`) replace inline dicts
- Algorithm citations reference academic papers
- HuggingFace API responses cached for 5 minutes
- Pagination bounds validation

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -50,7 +50,7 @@ All settings override via environment variables:
 - `LILBEE_VISION_MODEL` — vision OCR model (default: none)
 - `LILBEE_VISION_TIMEOUT` — per-page vision OCR timeout in seconds (default: `120`, `0` = no limit)
 - `LILBEE_LLM_PROVIDER` — backend: `auto` (default), `llama-cpp`, `ollama`, `openai`
-- `LILBEE_LLM_BASE_URL` — Ollama endpoint (default: `http://localhost:11434`)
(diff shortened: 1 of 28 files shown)


tobocop2:
Superseded by PR chain into release/next

PR #14, opened 2026-03-22
Retrieval quality phase 2: guardrails, context selection, architecture docs

tobocop2:
### Smarter query expansion
Query expansion now validates LLM-generated variants before using them. Each variant is checked for token overlap with the original question — if less than 30% of the original terms appear, the variant has drifted too far and is dropped. This prevents the LLM from generating search queries that have nothing to do with what you asked.

### Adaptive context selection
Instead of blindly sending the top-k results to the LLM, lilbee now picks chunks that maximize coverage of your question. If you ask "compare X and Y", it ensures both X and Y are represented in the context, not just the most relevant chunks about X.

### Architecture documentation
New docs/architecture.md with:
- Full system overview with Mermaid diagrams
- Every retrieval technique explained for humans with academic citations
- Configuration reference with defaults, rationale, and caveats for every setting
- Provider abstraction and ingestion pipeline documented

### Code cleanup
- Reasoning parser refactored from nested function to _TagParser class
- Removed Ollama-to-catalog name mapping (show names as-is)
- Removed divider comments from model_manager
- All new config fields registered and documented

### Config fields added
expansion_guardrails, expansion_skip_threshold, expansion_skip_gap,
max_context_sources, hyde, hyde_weight, reranker_model, rerank_candidates,
temporal_filtering — all with documented defaults and env var overrides.

### Academic references
- Expansion guardrails: standard token overlap validation
- Context selection: greedy set-cover approximation
- Adaptive threshold: grantflow (grantflow-ai/grantflow)
- MMR: Carbonell & Goldstein 1998
- HyDE: Gao et al. 2022
- Cross-encoder: Nogueira & Cho 2019

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,0 +1,301 @@
+# lilbee Architecture
+
+## What is lilbee?
+
+lilbee is a local knowledge base that lets you ask questions about your documents and get accurate, sourced answers. It runs entirely on your machine — no cloud, no API keys, no data leaving your computer.
+
+You point it at a folder of documents (markdown, code, PDFs, anything), it indexes them, and then you can search or chat with an AI that actually reads your files instead of making things up. Every answer comes with citations showing which documents it used.
+
+It works from the command line, as an API server, as an MCP tool for AI coding assistants, and via the REST API for any client integration.
+
+---
+
+## System Overview
+
+```mermaid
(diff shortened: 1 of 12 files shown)


tobocop2:
Superseded by PR chain into release/next

PR #27, opened 2026-03-27
test: RAG pipeline integration tests with real models

tobocop2:
17 integration tests using real GGUF models (Qwen3 0.6B + nomic-embed-text) via llama-cpp. no Ollama.

real embeddings, real vector search, real hybrid search, real MMR, real sync. LLM chat mocked only because 0.6B context window is too small for RAG prompts — but the search results feeding the prompt are fully real.

tests: keyword search, semantic search, MMR diversity across sources, per-source cap, code search with line numbers, concept graph boost, sync idempotency, delete removes from index, empty queries.

marked @pytest.mark.slow, excluded from make check by default. ~2.5 min on first run (model download), ~30s subsequent.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,8 @@ dev = [
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 asyncio_mode = "auto"
+markers = ["slow: integration tests that download and run real models"]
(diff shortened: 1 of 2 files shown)


tobocop2:
Merged directly to release/next

PR #28, opened 2026-03-27
test: interactive cli integration tests

tobocop2:
15 CLI integration tests using real llama-cpp inference. no mocked embeddings or LLM.

real nomic-embed-text embeddings, real Qwen3 0.6B chat, real crawl4ai against httpserver, real LanceDB, real kreuzberg + tree-sitter chunking. only concession: serialized embedding calls (llama-cpp-python is not thread-safe).

tests: keyword search, semantic search, ask with known facts, citations, MMR diversity, add/delete/sync/rebuild, code search with line numbers, web crawl then search, status JSON, version, empty db, init.

models auto-download from HuggingFace on first run (~800MB), cached in ~/.lilbee/models/. ~20s on subsequent runs.

### DIFF
--- a/tests/test_interactive_integration.py
+++ b/tests/test_interactive_integration.py
@@ -1,0 +1,422 @@
+"""Real integration tests with actual models -- no mocked embeddings or LLM.
+
+Exercises the full CLI pipeline with real llama-cpp inference:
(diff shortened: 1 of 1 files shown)


tobocop2:
Merged directly to release/next

PR #31, opened 2026-03-27
refactor: Lilbee class as single entry point with proper DI

tobocop2:
all pipeline operations go through the Lilbee class. no more public module-level functions for search/ask/sync. proper dependency injection throughout.

- query.py: public functions removed, private _search_context/_ask_raw/_ask_stream accept explicit config and provider
- api.py: Lilbee methods call private functions with explicit config/provider
- cli: get_bee() singleton, all commands go through it
- server: _get_bee() singleton, handlers receive Lilbee instance
- mcp: _get_bee() singleton, tools go through it
- tui: uses get_bee() from cli app

no compatibility shims. old function names are private. tests updated to match.

### DIFF
--- a/src/lilbee/api.py
+++ b/src/lilbee/api.py
@@ -1,8 +1,4 @@
-"""Programmatic access to lilbee's retrieval pipeline.
-
-Retrieval only -- no LLM chat. Search your indexed documents from Python.
-Optional features (concept graph, reranker) activate automatically when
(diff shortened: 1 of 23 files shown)


tobocop2:
Closing — replaced by composition architecture refactor (no get_bee globals, no _di suffixes, no swap hacks)

PR #35, opened 2026-03-28
feat: PATCH /api/config and PUT /api/models/embedding

tobocop2:
Adds two new endpoints so  consumers can update server config and switch embedding models.

**PATCH /api/config** — partial config updates.

**PUT /api/models/embedding** — switch the active embedding model, same pattern as the existing chat/vision endpoints.


tobocop2:
This should be rebased onto feature/model-management-v2 when ready to integrate. The PATCH /api/config and PUT /api/models/embedding endpoints are useful but need to target the active branch, not main.

tobocop2:
Merged into feature/tui-fixes-and-models (PR #37)

PR #36, opened 2026-03-29
feat: dynamic external model discovery

tobocop2:
Adds GET /api/models/external that queries the configured provider for available models using the stored API key.

Detects auth style from the base URL — Bearer token (standard) or X-Api-Key header. Parses the provider's /v1/models response to return a flat list of model IDs.

60-second cache that auto-invalidates when the base URL or API key changes. Generic error messages in responses (provider details only logged server-side).

1687 tests, 100% coverage.


tobocop2:
Superseded by feature/tui-fixes-and-models (PR #37). All external model work is now in the active branch.

PR #76, opened 2026-04-11
Shrink the setup wizard model cards so they fit on one screen

tobocop2:
The first-run setup wizard rendered each model card at roughly 113 rows tall, so a single row of cards blew through the viewport. Picking a model meant scrolling through large empty boxes before you could see what was on offer.

The card layout now travels with the ModelCard widget itself instead of living in the catalog screen's stylesheet, so the wizard and the catalog both show the cards at their intended compact size.

### DIFF
--- a/src/lilbee/cli/tui/screens/catalog.tcss
+++ b/src/lilbee/cli/tui/screens/catalog.tcss
@@ -61,48 +61,6 @@ GridSelect {
     }
 }
 
-ModelCard {
-    height: auto;
-    border: tall $surface-lighten-2;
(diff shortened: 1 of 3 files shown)


tobocop2:
Duplicate of #75 — both PRs landed the same architectural fix (move ModelCard styling into its DEFAULT_CSS so the widget is self-contained) within a few minutes of each other due to a team-internal task claim race. #75 was opened first and keeps the canonical branch. Thanks to both fixers for the independent confirmation that this is the right approach.

PR #106, opened 2026-04-14
Fix circular import crash when setting model via env var

tobocop2:
## Summary
- LILBEE_EMBEDDING_MODEL or LILBEE_CHAT_MODEL env vars caused ImportError on startup
- _normalize_model_tag validator imported from registry during Config init, triggering config -> registry -> models -> config circular import
- Inlined the DEFAULT_TAG constant ("latest") to break the cycle


tobocop2:
Superseded by BEE-4dj (ModelRef translation layer) which replaces _normalize_model_tag with parse_model_ref, handling both the circular import and API model name issues.

PR #124, opened 2026-04-15
fix: mark ask endpoints as read-only to fix 401 auth error

tobocop2:
POST /api/ask and POST /api/ask/stream returned 401 even with a valid session token. Both endpoints are semantically read-only (they query the knowledge base and generate answers, with no side effects), but they were missing the @read_only decorator that search already had. Without it, the auth middleware required a bearer token that the plugin doesn't send for query endpoints.

Added @read_only to ask_route and ask_stream_route, matching the pattern already used by search_route. Added two tests in TestAuthRequiredRoutes verifying both ask endpoints accept unauthenticated requests when auth is enabled. Also fixed a pre-existing import sort issue in tests/conftest.py.

### DIFF
--- a/src/lilbee/server/routes/search.py
+++ b/src/lilbee/server/routes/search.py
@@ -35,6 +35,7 @@ async def search_route(
 
 
 @post("/api/ask")
+@read_only
 async def ask_route(data: AskRequest) -> AskResponse:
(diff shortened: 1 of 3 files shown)


tobocop2:
Wrong fix — all HTTP endpoints should require auth. The bug is in the plugin token handling, not the server auth middleware.

PR #159, opened 2026-04-23
Gate file hashing on mtime vs ingested_at

tobocop2:
Sync previously ran SHA-256 on every discovered file every cycle, even when nothing had changed. At 10k files / 1GB that's a ~1s floor per sync.

Adds `_file_unchanged_since_ingest`: if the file's mtime is strictly older than the stored `ingested_at`, content can't have changed and the hash is skipped. Any failure (missing timestamp, unparseable ISO, stat error, naive datetime) falls through to the existing authoritative hash compare, so correctness is unchanged.

No schema change — reuses the `ingested_at` field that `upsert_source` already writes.

Refs bb-7yyf.

### DIFF
--- a/src/lilbee/ingest.py
+++ b/src/lilbee/ingest.py
@@ -10,6 +10,7 @@ import os
 import threading
 from collections.abc import Callable, Generator
 from dataclasses import dataclass
+from datetime import UTC, datetime
 from enum import StrEnum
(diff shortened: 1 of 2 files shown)


tobocop2:
Superseded by bundled complexity-review PR (coming soon).

PR #162, opened 2026-04-23
Add pytest-timeout guardrail for CI diagnosis

tobocop2:
## Why

Integration CI silently wedges for the full 90-minute runner budget with no stack trace or failing test name. Without a per-test timeout a single hung test eats the whole job and leaves zero diagnostic signal.

## What

Wire `pytest-timeout` into the test harness:

- 60s default for unit tests (well above the slowest legitimate case).
- 180s for anything under `tests/integration/` since real inference wants real headroom.
- Thread-based timeout method so Windows CI also dumps per-thread stacks on fire.

Individual tests can still opt into a higher cap with `@pytest.mark.timeout(X)` when they genuinely need it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,6 +79,7 @@ dev = [
     "types-pyyaml>=6.0.12.20250915",
     "pytest-xdist>=3.8.0",
     "pytest-forked>=1.6.0",
+    "pytest-timeout>=2.3",
 ]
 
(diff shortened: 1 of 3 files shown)


tobocop2:
Superseded by #163, which contains this commit plus the full fix (asyncio loop migration + wiki opt-out in non-wiki fixtures).

PR #163, opened 2026-04-23
Run integration sync on the shared session loop

tobocop2:
## Why

Integration CI wedges on `test_ingest_realistic_length_document` on every Linux and Windows matrix job, yet the same suite passes locally on Metal in seconds. Cause: a handful of tests spin up fresh event loops inside the test body, which tears down plumbing that the llama-cpp background workers depend on. Subsequent awaits deadlock silently, and on CPU-only runners the race window is wide enough to hit every time.

## What

Expose the session-scoped event loop through a new `run_async` fixture and migrate every test-body call that was reaching for `asyncio.run(...)`. The interactive test module's sync helpers move from module-level functions to fixtures so they share the same loop. No production code changes.

Stacked on the `pytest-timeout` PR so that if any test still exceeds 180s on CI the guardrail names the offender with a stack trace instead of another silent wedge.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,6 +79,7 @@ dev = [
     "types-pyyaml>=6.0.12.20250915",
     "pytest-xdist>=3.8.0",
     "pytest-forked>=1.6.0",
+    "pytest-timeout>=2.3",
 ]
 
 [tool.pytest.ini_options]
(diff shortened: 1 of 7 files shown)


tobocop2:
Squash-merged into feature/wiki-concepts. Integration matrix green on Linux + Windows; macOS 3.11 integration was still finishing at merge time (not gating per standing convention).

PR #165, opened 2026-04-23
Stop writing garbage wiki filenames

tobocop2:
## Problem

Running the wiki builder on real documents produced pages with names like `| | designer.md`, `158 vehicle.md`, `-answers.md`, `70-l.md`. These are not useful topics — they're table-markup fragments, PDF page numbers, and link-text residue that survived extraction. Worse, the generated H1 inside each page echoed the same garbage, so a reader saw `# | | designer` at the top of what was supposed to be a biographical article about a GM designer named Irv Rybicki.

## What changed

A sanity gate now rejects a label before it can become a slug or reach the language model as a topic. Labels need a real leading letter, at least three characters, mostly alphanumeric content, and no markdown-structural characters. The slug generator stops producing filenames that start with a stray hyphen. The prompt template receives a cleaned display name so the model never sees raw noise in the first place.

## Behavior

On the vehicle-manual and Chevrolet Caprice Wikipedia test corpora, the pages that previously landed as noise are no longer produced. The legitimate pages (`chevrolet.md`, `caprice.md`) are unaffected.

## Stack

First of seven PRs in a wiki-quality pass. Subsequent PRs continue to tighten upstream stages.

### DIFF
--- a/src/lilbee/concepts.py
+++ b/src/lilbee/concepts.py
@@ -92,6 +92,11 @@ def _filter_noun_chunks(doc: Any, max_concepts: int) -> list[str]:
     delimiters, page-number-prefixed tokens, sub-three-char fragments)
     never enter the co-occurrence graph and therefore never become a
     synthesis-page cluster label.
+
+    The gate runs on the lowercased form here while the NER extractor
+    gates on the original-cased surface; the two decisions match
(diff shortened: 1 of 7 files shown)


tobocop2:

PR #166, opened 2026-04-23
Don't feed table markup and page numbers to the NER model

tobocop2:
## Problem

The wiki entity extractor was handing kreuzberg's raw markdown output directly to spaCy. That included Wikipedia table rows (`| Designer | Irv Rybicki |`), standalone page numbers from PDF extraction (`42`), and CMS navigation chrome (`Edit this page`, `Jump to search`). spaCy dutifully ran named-entity recognition over all of it, turning markup scaffolding into candidate entities. The previous PR caught the resulting garbage at the slug stage, but the upstream noise still consumed extraction budget and cluttered the debug traces.

## What changed

Adds a small pre-clean pass that strips whole-line table rows, page-number-only lines, and a short list of common nav chrome strings before the text reaches spaCy. Normal prose is unchanged byte-for-byte; `Chapter 42 introduces` stays intact because only bare page-number lines are stripped.

## Behavior

On the same Wikipedia and PDF test corpora, the candidate-entity count per chunk drops substantially and the debug trace stops showing phantom entities derived from markup. The final page output is unchanged — the previous PR already blocked those labels from becoming pages. This one reduces wasted work upstream.

### DIFF
--- a/src/lilbee/wiki/entity_extractor/ner_concepts.py
+++ b/src/lilbee/wiki/entity_extractor/ner_concepts.py
@@ -25,10 +25,40 @@ _ALLOWED_NER_LABELS: frozenset[str] = frozenset(
 )
 _WHITESPACE_RE = re.compile(r"\s+")
 
+# Pre-spaCy markdown-noise strippers. Compiled once at module scope so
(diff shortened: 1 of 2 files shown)


tobocop2:

PR #167, opened 2026-04-23
Filter numeric NER tags out of wiki candidates, make the pipeline debuggable

tobocop2:
## Problem

spaCy's named-entity recognizer tags chunks of text with dozens of categories, many of which aren't useful wiki topics: `QUANTITY`, `CARDINAL`, `DATE`, `TIME`, `MONEY`, `PERCENT`, `ORDINAL`. A number like `42 miles` or `$199` would sail through the previous filters and become a wiki page candidate. The hardcoded allowlist was also missing two categories that do belong (`FAC` for facilities and `NORP` for nationalities).

Separately, when the extractor ran there was no way to see where entities were being dropped. Silent filters make pipeline debugging painful.

## What changed

The allowlist moves onto the config dataclass with replace-semantics: narrowing to `PERSON,ORG` does exactly that, not "the defaults plus those two". `FAC` and `NORP` join the default set. The prior hardcoded list of seven categories is gone.

A single DEBUG line now fires once per extraction pass with the full funnel — raw entity count, raw noun-chunk count, drops by type filter, drops by label sanity (split by kind), surfaces kept on each side. Built only when DEBUG logging is active so there's no cost in production.

## Behavior

Running the extractor with `LILBEE_LOG_LEVEL=DEBUG` now shows the reduction stages. On the test corpora, roughly 40% of raw spaCy output gets filtered at the type stage — numbers and dates that used to slip through.

### DIFF
--- a/src/lilbee/config.py
+++ b/src/lilbee/config.py
@@ -879,12 +879,12 @@ class Config(BaseSettings):
     @field_validator("concept_allowed_ent_types", mode="before")
     @classmethod
     def _parse_ent_types(cls, v: Any) -> frozenset[str]:
-        """Replace-semantics override, unlike ``ignore_dirs``.
-
-        A user narrowing the NER type set (e.g. ``PERSON,ORG`` only)
-        wants exactly those kinds, not the project defaults unioned
-        with the override. Accepts comma-separated strings from env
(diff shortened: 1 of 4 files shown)


tobocop2:

PR #168, opened 2026-04-23
Review, accept, or reject wiki drafts from the command line

tobocop2:
## Problem

The wiki builder already knew how to send a generated page to `drafts/` when it drifted too far from the prior version or scored below the faithfulness threshold. It just never gave the user a way to look at those drafts afterward. They piled up invisibly, there was no diff against the current published page, no way to accept a draft as the new published version, and no way to reject and delete one. "Drafts" was a dead-end directory.

## What changed

Four new commands under `lilbee wiki drafts`:

- **`list`** shows every pending draft with its drift percent, faithfulness score, and whether a published counterpart exists.
- **`diff <slug>`** prints a unified diff against the published page (or shows the draft as all-new when there's nothing published yet).
- **`accept <slug>`** replaces the published page with the draft and re-indexes its chunks so it shows up in search. The sequence is deliberately write-then-reindex-then-unlink: if re-indexing fails, the draft stays on disk and the user can retry.
- **`reject <slug>`** deletes the draft without touching the published page or the index.

The first two operations are also exposed through MCP; accept and reject stay CLI-only so destructive actions need explicit intent.

## Behavior

After a drift or a low-faithfulness event, `wiki drafts list` shows the pending pages. The user diffs, decides, and moves on. Nothing accumulates silently anymore.

### DIFF
--- a/src/lilbee/cli/commands.py
+++ b/src/lilbee/cli/commands.py
@@ -1260,3 +1260,134 @@ def wiki_update(
     lands in the ingest-hook task and will re-route this command then.
     """
     wiki_build(data_dir=data_dir, use_global=use_global)
+
+
+drafts_app = typer.Typer(help="Review wiki drafts: list, diff, accept, reject.")
+wiki_app.add_typer(drafts_app, name="drafts")
+
+
+@drafts_app.command(name="list")
+def wiki_drafts_list(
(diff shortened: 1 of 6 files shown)


tobocop2:

PR #169, opened 2026-04-23
Preview a wiki build without spending an LLM call, and trace pages back to their sources

tobocop2:
## Problem

A wiki build that produces garbage takes the same amount of time as one that produces useful output. If the extraction pipeline emits bad candidate entities — because of a new corpus, a config mistake, a noisy PDF — you only find out after the language model has burned through all of them. On an eight-minute build against a 20-page PDF, that's eight minutes of wasted tokens before anyone sees the list.

Separately, once a wiki page lands it's hard to trace back to the chunks that produced it. If a page looks wrong, you have no way to audit which source chunks fed the generation without re-running the whole pipeline.

## What changed

**Dry-run.** `lilbee wiki build --dry-run` runs the full extraction pipeline — NER, markdown clean, type filter, label sanity — and prints a table of candidate entities with their mention counts and source lists. No LLM call happens. The QA loop on the test corpus dropped from roughly eight minutes to two seconds.

**Provenance.** Every generated page now carries a `provenance` block in its frontmatter naming the chunks it was generated from (source file and chunk index) and the extraction method used. A bad page is auditable after the fact without re-running anything.

## Behavior

Pre-existing wiki pages without the provenance block still parse — the new field is additive and optional.

### DIFF
--- a/src/lilbee/cli/commands.py
+++ b/src/lilbee/cli/commands.py
@@ -1201,6 +1201,11 @@ def _fail_wiki_disabled() -> None:
 def wiki_build(
     data_dir: Path | None = data_dir_option,
     use_global: bool = global_option,
+    dry_run: bool = typer.Option(
+        False,
+        "--dry-run",
+        help="Run extraction only; skip every LLM call. Prints the candidate entities.",
+    ),
 ) -> None:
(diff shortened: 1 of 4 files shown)


tobocop2:

PR #170, opened 2026-04-23
Catch garbage page titles before they ship, without another LLM call

tobocop2:
## Problem

The faithfulness check grades each generated page on how well its prose reflects the source chunks. The problem is that a page with a nonsense title can still score high if the body text is internally coherent. A real example from the test corpus: a page titled `# | | designer` with perfectly readable prose about an automotive designer scored 0.90 and shipped as-is. The faithfulness check doesn't look at titles — only at whether the body is grounded in the sources.

## What changed

Before the language model scores faithfulness, a cheap deterministic check now runs over the generated page. Three invariants:

1. The H1 heading must itself pass the label-sanity gate — no pipes, hashes, angle brackets, or leading non-alphabetic characters.
2. The cleaned display name of the concept must appear in the heading.
3. The body must mention the concept at least once outside the heading.

If any of those fail, the faithfulness score is capped at 0.5 regardless of what the language model says. With the default threshold at 0.7, that cap automatically routes the page to drafts for human review instead of publishing it.

## Behavior

No extra LLM call. The coherence check is pure text comparison, runs in microseconds, and only affects pages that were going to land with nonsense titles anyway.

### DIFF
--- a/src/lilbee/wiki/gen.py
+++ b/src/lilbee/wiki/gen.py
@@ -57,6 +57,7 @@ from lilbee.wiki.shared import (
     WIKI_LOG_ACTION_GENERATED,
     PageTarget,
     clean_label_for_display,
+    is_valid_label,
     make_slug,
     parse_frontmatter,
 )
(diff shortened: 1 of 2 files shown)


tobocop2:

PR #171, opened 2026-04-23
Cut wiki build LLM calls by ~95% and fix concept pages at the source

tobocop2:
## Problem

Two structural issues in the wiki builder surfaced when running it against a realistic 20-page corpus with a 8-billion-parameter local model:

**It's far too slow.** Every wiki page cost two language-model calls — one to write the page and one to score how faithfully it reflects the sources. A corpus that produces fifty candidates means a hundred calls, thirty to sixty minutes of wall time, and the builder had no way to parallelize them because local inference runs on a single GPU.

**Most of those pages are noise.** The concept-page pipeline was driven by spaCy's noun-chunk output — every grammatical noun phrase became a candidate. That means "brake systems" and "the vehicle" and "you" and "cro" (a column-wrap fragment from a PDF) all competed for a page. Filter tuning couldn't distinguish them because they're all syntactically noun phrases. The only honest fix is to change the extractor.

## What changed

Three coordinated changes that together turn 38 LLM calls into 2 on the test corpus.

### Scoring faithfulness without an LLM call

The faithfulness check used to ask the model how well a page reflected the source chunks. That's one call per page. It's now a cosine similarity between the page's embedding and the mean of the source chunks' embeddings — both of which the system already has. The deterministic title/body coherence gate from the previous PR still runs first as a hard filter. Halves the call count on its own.

### Concept pages now come from the model's judgment, not spaCy

The noun-chunk extraction path for concept pages is gone. In its place, the new batched generation pass (below) asks the model to identify three to five concepts worth a page from each source's content — using its actual semantic understanding of what a useful wiki topic is. Pages like `the-vehicle.md`, `warning.md`, `cro.md` disappear; pages like `brake-systems.md` become possible because the model can tell those apart. Existing noun-chunk-derived pages on disk are archived once, and any `[[link]]` references to them are unwrapped.

### One LLM call per source document

The builder now groups extracted entities by their primary source and issues a single call per source that both identifies concepts and writes a section for every entity and every identified concept at once. One source, one call, many pages. Sections that fail to parse — or that collide on the same slug across two sources — route to the drafts review surface as PENDING markers so nothing is lost silently.

## What the user sees

- Build on a 20-page test corpus: from ~38 LLM calls and roughly 15 minutes wall time, to 2 calls and under 3 minutes.
- Concept pages have real topics, generated from the model's judgment with citations back to the source chunks.
- A preview (`wiki build --dry-run`) completes in seconds and shows the candidate entities before any tokens are spent.
- Pending markers for parse failures or slug collisions show up in `wiki drafts list` and can be resolved with `accept` or `reject`.

The concept graph that powers search-result boosting and cross-source synthesis pages is untouched. Only the per-page concept generation changed source.

## Stack

Final PR in a seven-PR arc tightening the wiki builder's output quality and throughput. Stacks on the preceding six.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -324,12 +324,15 @@ entity pages that compound across sources rather than one page per
 document. The layout under `$data_root/$wiki_dir/` is:
 
 ```
-concepts/    one page per noun-phrase topic (e.g. braking-systems.md)
+concepts/    one page per LLM-curated concept from the source (e.g. braking-systems.md)
 entities/    one page per proper-noun entity (e.g. henry-ford.md)
 summaries/   legacy per-source pages (still supported, not the default)
 synthesis/   cross-source pages produced by `wiki synthesize`
-drafts/      low-faithfulness drafts that missed the threshold
-archive/     pages retired by `wiki prune`
+drafts/      low-faithfulness drafts, drift drafts, and PENDING markers
+             (parse-failure or slug-collision) surfaced via `wiki drafts list`
+archive/     pages retired by `wiki prune` (plus the one-time Phase D
+             migration archive at `archive/concepts/` from the pre-Phase-D
+             noun-chunk generator)
 index.md     auto-generated table of contents, grouped by page type
 log.md       append-only audit trail (## [YYYY-MM-DD HH:MM] op | details)
 ```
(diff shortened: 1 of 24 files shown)


tobocop2:

PR #174, opened 2026-04-23
TUI startup crash: stop overwriting the configured model with the Select widget's default

tobocop2:
## Problem

Launching `lilbee chat` was crashing on startup on any data directory where the alphabetically-first installed model wasn't one of lilbee's featured models. Users got a validation error instead of the chat screen.

## What changed

The TUI starts cleanly again. The model-picker no longer overwrites the user's configured model with the Select widget's auto-picked default on startup, so a machine that has a non-featured model installed (even one that isn't the active one) doesn't block the user from using the TUI any more.

## Verification

Two regression tests cover the startup path that previously overwrote config.

### DIFF
--- a/src/lilbee/cli/tui/widgets/model_bar.py
+++ b/src/lilbee/cli/tui/widgets/model_bar.py
@@ -324,9 +324,18 @@ class ModelBar(Widget, can_focus=False):
 
     @on(Select.Changed, "#chat-model-select")
     def _on_chat_model_changed(self, event: Select.Changed) -> None:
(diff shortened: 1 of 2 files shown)


tobocop2:
Superseded: equivalent fix merged into feature/wiki-concepts as commits a3c029b..0d10636, with stronger synchronous stale-event defense (str(event.value) != str(sel.value)) replacing the Windows-fragile call_after_refresh approach. Also lands a real async-drain regression test (test_populate_survives_auto_pick_race).

PR #258, opened 2026-05-16
demos: tui-setup with Hide 1000s

tobocop2:
Bumps Hide to 1000s (~2.4x the empirical 419s done time). Downloads are guaranteed to be at done before the question fires.


tobocop2:
Superseded by #261 (merged).

PR #259, opened 2026-05-16
demos: tui-setup re-render (Hide 1000s)

tobocop2:
Re-render with the corrected Hide. Question fires only after both downloads finish.


tobocop2:
Superseded by #262 (merged).

PR #267, opened 2026-05-17
Multi-GPU model serving and agent integration (opencode + hermes) 

### LINKED ISSUE #473: gguf-parser is built from unpinned upstream main; schema rename breaks the VRAM estimator
A fresh engine build produces a gguf-parser whose output our VRAM estimator cannot read, and the resulting error message points at the wrong subsystem entirely. The build is broken because it tracks a moving upstream branch instead of a pinned release.

## What broke

`tools/wheel-build/build_llama_server.sh` clones gguf-parser-go at `_GGUF_PARSER_REF="main"`. Upstream renamed the estimate payload in its JSON output (`estimate.items` became `estimate.memory`) sometime after v0.24.1. Any engine built after that rename returns JSON that `_parse_estimate` in `providers/fleet/vram.py` cannot read, so every model fails sizing and the fleet planner starts nothing.

## Why it was hard to diagnose

The real error never surfaces. Sizing raises `ProviderError`, the planner's `except` swallows it, and every role gets skipped with a warning that blames a different subsystem:

```mermaid
flowchart LR
    build["engine built today<br/>from upstream main"] --> newjson["gguf-parser emits<br/>estimate.memory"]
    newjson --> parse["_parse_estimate expects<br/>estimate.items: fails"]
    parse --> swallow["planner catches the error<br/>and skips the role"]
    swallow --> lie["log says: model is<br/>not installed"]
    lie --> hunt["you debug the model registry,<br/>which was fine all along"]
```

The models were installed and resolvable the whole time; the warning sends you into the model registry when the estimator is what died.

## How we know

- Hit on a fresh pod build on 2026-07-03. The same branch worked a day earlier only because that pod's engine binary was built before the rename.
- A binary built from the v0.24.1 tag emits the old `items` schema and works, which brackets the upstream change to main within the last few weeks.

## Blast radius

Not just source builds: the `lilbee-engine` wheel that `uv sync` installed on the same pod also bundles a memory-schema parser, so recently published wheels likely carry the broken binary and should be audited.

## Suggested fix

- Pin `_GGUF_PARSER_REF` to v0.24.1, the last tag with the `items` schema. Never build a release artifact from a moving branch.
- Audit and republish any `lilbee-engine` wheels built after the upstream rename.
- Teach `_parse_estimate` to accept both shapes, or fail loudly with the actual parse error instead of the misleading not-installed warning, so the next schema change is a one-line log read instead of a debugging session.

tobocop2:
## Problem

The local model engine work has lived on feat/local-model-api while main carried the 0.6.66 beta line. The branch is done: the llama-server fleet serves every role, multi-GPU placement is hardened and QA'd, and the remaining bug backlog on the branch is fixed and closed.

## Solution

Bring the whole line to main. Highlights since the branches diverged:

- A managed llama-server fleet (via llama-swap + gguf-parser) replaces llama-cpp-python for chat, embeddings, rerank, and vision OCR, with multi-GPU placement: auto bin-packing with usable-context floors, tensor-split across cards, an interactive TUI placement editor, and placement surfaces across CLI, HTTP, and MCP sharing one serializer.
- Agent integrations (opencode, hermes) with a shared-home launcher, streaming and tool-calling fixes for reasoning models, and cross-transport parity sweeps.
- The final bug batch: placement editor and drawer fixes, chat submits held through fleet reloads, vision mmproj recovery on pull, network-volume and no-NVLink guardrails validated on live pods, and the gguf-parser estimator accepting both upstream schemas (closes #473).
- Main's beta-line fixes are preserved through the merge: dependency caps with guard tests and the per-cell release wheel attach. The CUDA-runtime bundling for the old wheel pipeline is superseded by the engine's cuda12 extra.

8412 tests at 100% coverage on the merged tree.


tobocop2:
Post-audit CI: green for everything this PR touches.

5 iterations of audits (AGENTS.md code-smell sweep + cross-cutting-api-change + blast-radius-review), every finding addressed:
- Pydantic models for the /v1 wire boundary (request/response/stream chunks/models list); litestar parses + validates, no hand-rolled `payload.get("...")` left in the route layer
- FinishReason and ToolChoiceMode StrEnums replace inline `Literal` sets
- temperature / top_p / top_k / max_tokens get pydantic range validators
- Cross-module private API leak fixed (`_server_session` / `_chat_model_refs` -> `running_server_session` / `installed_chat_model_refs`)
- Dead `CONTEXT_LENGTH_EXCEEDED` enum value removed
- 64-line stream chunker split into a `_StreamMapper` helper
- 49-line `opencode_cmd` shed two helpers
- Nuitka binary build now bundles `src/lilbee/skills/` so the standalone executable ships the skill (wheel already did)

Remaining failure (pre-existing): `test (windows-latest, 3.12)` on `test_populate_library_list_with_only_frontier_rows` -- TUI catalog test, no diff overlap, flagged in earlier comments.

tobocop2:
Final CI: green except for the pre-existing windows-3.12 TUI catalog flake (`test_populate_library_list_with_only_frontier_rows`). It's the same flake flagged in earlier comments, present on main, no diff overlap with this PR.

Validated end-to-end on macOS with real opencode 1.14.50: `lilbee launch opencode` correctly spawns a fresh server (or reuses a healthy one), wires the inline config + skill + picker state, and opencode connects to and queries the lilbee provider. The launcher now health-probes before reusing an existing session so a stale `server.port` file from a crashed prior server doesn't trap opencode in a reconnect loop.

PR #285, opened 2026-05-24
Multi-GPU inference via a managed llama-server fleet (opt-in)

tobocop2:
## Problem

On a multi-GPU machine, lilbee ran all inference in a single in-process worker, one request at a time. Extra GPUs sat idle, and concurrent requests queued behind one model.

## Solution

An opt-in mode that runs inference on a fleet of llama.cpp `llama-server` processes spread across the GPUs. lilbee places the models, supervises the processes, and routes each request to the least-busy one.

- **Off by default.** Nothing changes, installs, or downloads unless you opt in: `pip install lilbee[multi-gpu]`, then pick the `multi-gpu` provider (TUI settings, MCP, HTTP, or `LILBEE_LLM_PROVIDER`).
- **Uses every GPU.** Models pack across cards by VRAM; a model too big for one card is split proportionally to each card's capacity, and small models share a card.
- **What runs where.** All four roles run on the fleet, each mirroring the in-process primitive over HTTP so robustness is unchanged: chat/vision via `/v1/chat/completions` (vision adds an `--mmproj` projector), embed/rerank via `/v1/embeddings` (rerank with `--pooling rank` and the same `query</s></s>candidate` pairing as in-process — never the template-dependent `/v1/rerank`). A model override, or a vision model with no resolvable mmproj, falls back to in-process. PDF rasterizes pages locally, then OCRs on the vision server.
- **Self-contained binary.** lilbee builds `llama-server` itself (CUDA / Vulkan / ROCm / Metal) and ships it as the opt-in extra, plus a standalone binary for channels that can't pip-install. These builds run on their own CI path and never block a release.
- **Already have a cluster?** Point the existing `remote` provider at it instead.

## Architecture

```mermaid
flowchart TD
    APP["App<br/>chat · embed · rerank · vision · search · ingest"]
    FP["FleetProvider"]
    LOCAL["in-process llama-cpp<br/>(fallback)"]
    SUP["Planner + supervisor<br/>VRAM bin-pack · pin per backend<br/>health · restart · reap orphans"]

    APP --> FP
    FP -->|"least-busy healthy server"| FLEET
    FP -. "override / no server" .-> LOCAL

    subgraph FLEET["Managed llama-server fleet"]
        direction LR
        CS["chat / vision<br/>(/v1/chat/completions)"]
        ES["embed / rerank<br/>(/v1/embeddings)"]
    end

    SUP -. "spawn · pin · health · restart" .-> FLEET

    CS --> G0["GPU 0"]
    CS --> G1["GPU 1"]
    ES --> G1
```

## Built to be safe under concurrency and across cards

The fleet is the concurrent, multi-card path, so it is hardened for exactly that:

- **No device-index crossing.** GPUs are enumerated and pinned through the binary's own `llama-server --list-devices`, per backend (CUDA with `CUDA_DEVICE_ORDER=PCI_BUS_ID`, ROCm via `ROCR/HIP_VISIBLE_DEVICES`, Vulkan via `GGML_VK_VISIBLE_DEVICES`) — never a Vulkan index fed to CUDA.
- **No request-routing races.** Single-flight fleet build, atomic in-flight counter, health-aware least-in-flight routing.
- **No stranded GPUs.** Own process group per server, ports claimed at spawn, a crashed parent's orphans reaped on next start, and a monitor that restarts a dead server with backoff.
- **Hardware QA.** `tools/qa/multi_gpu_smoke.py` verifies all of the above on a real multi-GPU host (the mocked unit tests can't).

Design and placement details live in `docs/architecture.md`.

## Prior art / design references

The architecture follows the established practice in this space:

- llama.cpp [multi-gpu.md](https://github.com/ggml-org/llama.cpp/blob/master/docs/multi-gpu.md) — split modes (`layer` default; `row` deprecated; `tensor` experimental, needs flash-attn) and `--tensor-split`.
- llama.cpp [`--ctx-size` is total ÷ `--parallel` (#11681)](https://github.com/ggml-org/llama.cpp/issues/11681) — KV sizing.
- llama.cpp [`/health` returns 200 only when loaded (commit)](https://github.com/ggml-org/llama.cpp/blob/master/tools/server/README.md) — readiness gate.
- llama.cpp [#20085: `--reranking` + `--embedding` on one process returns zeros](https://github.com/ggml-org/llama.cpp/issues/20085) — the fleet reranks via rank-pooling `/v1/embeddings` (not `/v1/rerank`), so this doesn't apply.
- llama.cpp [Vulkan vs CUDA device index ordering differ (#12443)](https://github.com/ggml-org/llama.cpp/discussions/12443) and CUDA [`CUDA_DEVICE_ORDER`](https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/environment-variables.html) — the device-index hazard.
- [Ollama GPU scheduling/UUID pinning](https://docs.ollama.com/gpu) and [vLLM `gpu_memory_utilization`](https://docs.vllm.ai/en/stable/configuration/optimization/) — VRAM budgeting and the 90% headroom.
- [llama-swap](https://github.com/mostlygeek/llama-swap) — single-flight model startup, per-server concurrency cap, process lifecycle.
- [Ray Serve routing policies](https://docs.ray.io/en/latest/serve/llm/architecture/routing-policies.html) — least-loaded request routing.
- [HF TGI `--num-shard` per GPU-set](https://huggingface.co/docs/text-generation-inference/basic_tutorials/launcher) and [Triton `instance_group`](https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/user_guide/model_configuration.html) — one-server-per-device-set placement.
- [GPUStack gguf-parser](https://github.com/gpustack/gguf-parser-go) — GGUF-metadata VRAM estimation.


tobocop2:
Superseded by #297, which carries this entire multi-GPU/llama-server stack plus the migration that makes the managed llama-server fleet lilbee's sole local inference engine (the opt-in multi-GPU provider concept is gone; it's now the default that scales to N GPUs). Closing in favor of #297.

PR #287, opened 2026-05-24
Classify provider errors by type, not message strings

tobocop2:
## Problem

When a remote model call failed, the SDK path flattened every backend exception into a single `ProviderError` carrying only a stringified message. To tell *why* it failed (bad key vs. rate limit vs. prompt too long vs. model not found), the only option was to match substrings of that message. Those wordings are provider-specific and change between SDK versions, so any such matching is brittle and does not generalize across models.

## Solution

Errors are now classified by *type* into a small provider-agnostic set of kinds, and callers branch on the kind instead of the message. `ProviderError` carries a `kind`, and its string form appends an actionable hint for that kind, so the TUI, HTTP, and MCP surfaces all get the guidance without any per-surface wiring.

The litellm boundary does the classification by exception type. litellm already normalizes every backend's errors into one exception hierarchy, so the same mapping covers Gemini, OpenAI, Anthropic, Ollama, and the rest identically. The mapping walks the class hierarchy so a context-window error is recognized as exactly that rather than a generic bad request.

Errors that aren't recognized stay `unknown` and behave exactly as before, so nothing downstream changes for them. The local in-process engine is intentionally left as-is: its failures are raised inside a worker process and serialized to text across the transport, so they can't carry a kind without a much larger change.

### DIFF
--- a/src/lilbee/providers/base.py
+++ b/src/lilbee/providers/base.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from collections.abc import Callable, Iterator
+from enum import StrEnum
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, Protocol, TypeVar, overload, runtime_checkable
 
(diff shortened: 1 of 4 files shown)


tobocop2:
Folding this into #286, which keeps the type-based ProviderErrorKind classification from here plus the mid-stream cause unwrap, blob-free messages, embeddings/reranking coverage, and the server-stream fix. Superseded by #286.

PR #291, opened 2026-05-25
crawler: strip reference-style [[N]](url) links from crawled markdown

tobocop2:
## Problem

Crawled pages from reference-heavy sites (Wikipedia, etc.) render badly in Obsidian. crawl4ai converts a page's footnote markers (the superscript `[1][2][3]`) into nested-bracket Markdown links — `[[2]](https://…#cite_note-2)`. The inner brackets make it a valid link with the text `[2]`, but Obsidian treats `[[…]]` as its own wikilink syntax, so it parses `[[2]]` as a broken link to a note named "2" and renders the trailing `(url)` as literal text. A Wikipedia article carries hundreds of these, so the body fills with broken wikilinks and inline URLs.

## Solution

Collapse the nested brackets to single ones at save time (`[[N]](url)` → `[N](url)`) so the reference renders as an ordinary link. This is non-lossy — the link text and URL are preserved, and ordinary `[text](url)` links are untouched — so it's a strict rendering-correctness improvement for every crawled page, not a content edit. The content-hash dedup runs on the normalized text so re-crawls stay consistent. Added unit tests for the normalizer and the save path; `tests/test_crawler.py` is green (176).

### DIFF
--- a/src/lilbee/crawler/save.py
+++ b/src/lilbee/crawler/save.py
@@ -133,6 +133,23 @@ def content_hash(text: str) -> str:
     return hashlib.sha256(text.encode()).hexdigest()
 
 
+# Reference-style nested-bracket links, e.g. Wikipedia footnote markers like
+# ``[[1]](https://en.wikipedia.org/wiki/Foo#cite_note-1)``. The inner brackets
+# make this a normal Markdown link with the text ``[1]``, but readers that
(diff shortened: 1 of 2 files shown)


tobocop2:
Superseded by #293, which combines this with the first-run browser-setup visibility fix into one PR for easier testing.

PR #292, opened 2026-05-25
crawler: surface first-run browser setup in the Task Center

tobocop2:
## Problem

The first crawl on a fresh install stalls with no feedback. crawl4ai's one-time browser warmup (opening the browser the first time) can take many seconds, and the crawl flow emits no progress during it, so the Task Center sits blank and the crawl looks hung. Setup progress was only emitted when Chromium had to be *downloaded*; once it's present, the warmup is silent.

## Solution

Bracket the crawler open in `crawl_recursive` with `setup_start`/`setup_done` events (a "browser" component, no size estimate so it reads as indeterminate). The plugin already renders setup events as a "preparing crawler" stage, so the first crawl now shows that stage instead of a silent stall. The Chromium-download path keeps its own granular progress. Added a test that the warmup is bracketed by setup events; `tests/test_crawler.py` is green (174).

### DIFF
--- a/src/lilbee/crawler/runner.py
+++ b/src/lilbee/crawler/runner.py
@@ -37,7 +37,15 @@ from lilbee.runtime.progress import (
     CrawlStartEvent,
     DetailedProgressCallback,
     EventType,
+    SetupDoneEvent,
+    SetupStartEvent,
 )
+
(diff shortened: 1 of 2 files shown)


tobocop2:
Superseded by #293, which combines this with the reference-link normalization into one PR for easier testing.

PR #299, opened 2026-05-29
Surface cloud (API-key) and Ollama models in the HTTP model catalog

tobocop2:
## Problem
With a cloud API key configured — Gemini, OpenAI, Anthropic — none of those models showed up in the model catalog over the HTTP API, and neither did Ollama-managed models. The terminal UI listed them, but anything talking to the server over HTTP (the Obsidian plugin in particular) only ever saw local models.

## Solution
The catalog now includes cloud models, grouped by provider with a key-ready indicator, and Ollama models, alongside local ones. They come back ready to use with no download, since they run off-box. Selecting a cloud model as the chat model works now too — it was previously rejected as 'not available.' Discovery is cached briefly so browsing the catalog stays responsive.

### DIFF
--- a/src/lilbee/catalog/types.py
+++ b/src/lilbee/catalog/types.py
@@ -25,6 +25,8 @@ class ModelSource(StrEnum):
 
     NATIVE = "native"  # lilbee's GGUF files in cfg.models_dir
     REMOTE = "remote"  # Models managed by a remote SDK-backed service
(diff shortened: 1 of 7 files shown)


tobocop2:
Superseded by #302, which fuses this branch with the other two overlapping change sets (frontier/Ollama catalog, installed-source granularity, LM Studio) into one PR.

PR #300, opened 2026-05-29
Support LM Studio alongside Ollama

tobocop2:
## Problem

Ollama was hardcoded across the routing and discovery layers, so there was no clean way to support another local OpenAI-compatible server like LM Studio.

## Solution

Add a small local-server registry (Ollama, LM Studio) that drives URL detection, the litellm wire prefix, model listing, and read-only capability. Both servers work for chat, embedding, and rerank, and their models are listed from the catalog. Neither is pulled by lilbee: models are managed in their own app and surface here once present. lilbee downloads native GGUF models only.

The README and the base-URL setting now mention both.


tobocop2:
Superseded by #302, which fuses this branch with the other two overlapping change sets (frontier/Ollama catalog, installed-source granularity, LM Studio) into one PR.

PR #373, opened 2026-06-17
Make GPU ingest work on driver-only images by wiring the CUDA runtime

tobocop2:
## Problem

The fleet llama-server links the CUDA 12 runtime (`libcudart.so.12`, `libcublas.so.12`, `libnvrtc.so.12`), but driver-only GPU images (common on RunPod) ship only `libcuda.so` from the kernel driver. The runtime libraries are missing, so llama-server exits before binding its port and llama-swap surfaces only an opaque "exited prematurely" — every vision/embed call then returns nothing. The benchmark has been working around this by hand: pip-installing the CUDA runtime wheels and putting them on `LD_LIBRARY_PATH`.

## Solution

New `fleet/cuda_runtime.py`. On Linux, `llama_server_runtime_env()` now prepends the lib directories of any installed `nvidia-cuda-runtime-cu12` / `nvidia-cublas-cu12` / `nvidia-cuda-nvrtc-cu12` wheels to the spawned server's `LD_LIBRARY_PATH`, so GPU ingest works out of the box once those wheels are present. `plan_all_launches()` preflights the runtime with `ldd` before the device probe; when the libraries are still unresolved it raises an actionable error naming the exact pip packages to install, instead of letting the probe fail opaquely. A CPU/Vulkan build does not link the CUDA runtime, so the preflight is a no-op there, and it does nothing off Linux.

Unit tests cover the env wiring, the wheel-dir discovery, and the preflight (missing libs, clean libs, no ldd), plus the planning preflight-before-probe ordering. Full lint, format-check, mypy, and imports-check pass.

Closes bb-b5lr.

### DIFF
--- a/src/lilbee/providers/fleet/binary.py
+++ b/src/lilbee/providers/fleet/binary.py
@@ -84,11 +84,14 @@ def resolve_gguf_parser() -> Path:
 
 
 def llama_server_runtime_env() -> dict[str, str]:
-    """Extra environment for a spawned ``llama-server`` (empty on every platform).
+    """Extra environment for a spawned ``llama-server``.
 
     The bundled wheel ships its own ggml/llama/mtmd next to the binary with a baked
-    rpath (``@loader_path`` on macOS, ``$ORIGIN`` on Linux), and a bring-your-own
-    binary carries its own libraries, so the fleet never injects a library search
-    path. Kept as the single hook for any future per-spawn environment.
+    rpath (``@loader_path`` on macOS, ``$ORIGIN`` on Linux), but a CUDA build also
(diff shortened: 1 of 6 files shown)


tobocop2:
Scrapping this approach. The mechanism (CUDA runtime via the official nvidia pip wheels) is right, but the implementation is over-built for what is a packaging gap: ~123 lines including a runtime ldd-subprocess preflight to diagnose a missing dependency. Replacing with a minimal version: a cuda12 optional-dependency extra so the wheels are actually installed, plus a small LD_LIBRARY_PATH shim, and no preflight. Tracked under bb-b5lr.

PR #377, opened 2026-06-17
Add a GLM-4-9B chat-template override for tool calls

tobocop2:
## Problem

GLM-4-9B's GGUF-embedded chat template renders no tool-calling Jinja. The fleet's
`supports_tools` probe inspects the template, finds no tool tokens, and rejects
the model with "does not support tool calls" — so GLM-4-9B can never be offered
tools, even though the model itself can do them.

## Solution

A small `--chat-template-file` override registry, mirroring the existing
`resolve_ocr_prompt` pattern: it maps a family token to a packaged tool-capable
template, injects `--chat-template-file` for the chat role in `build_server_argv`,
and short-circuits `supports_tools` to True when an override exists. Only GLM-4-9B
is overridden — a pod run confirmed its override restores end-to-end tool calls.

Other families whose GGUF templates break tools (internlm2, olmo3, and similar)
need engine-level tool-call parsers, not a template, so they are intentionally
left to their own template here.

**Depends on the engine bump.** GLM-4-9B's override only produces structured tool
calls once the engine can parse GLM-4's tool output, which lands with the bundled
llama.cpp bump (#370). On the older engine the override changes a clean rejection
into a timeout, so this should merge after #370, not before.

### DIFF
--- a/src/lilbee/cli/commands/setup.py
+++ b/src/lilbee/cli/commands/setup.py
@@ -126,6 +126,7 @@ def _self_check_server(role: WorkerRole, model_path: Path) -> tuple[SwapManager,
         llama_server_runtime_env,
         resolve_llama_server,
     )
+    from lilbee.providers.fleet.chat_templates import resolve_chat_template
     from lilbee.providers.fleet.client import LlamaServerClient
     from lilbee.providers.fleet.launch import InstanceLaunch
     from lilbee.providers.fleet.swap_manager import SwapManager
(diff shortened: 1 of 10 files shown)


tobocop2:
Dropping glm-4-9b support rather than shipping a special case for it. Today's multi-tool pod run showed it passes a single tool call but times out under two sequential calls (it's on llama.cpp's generic parser path, no native GLM-4 parser). Combined with: it's an old (mid-2024) model superseded by newer GLM releases, it's the only consumer of the chat-template-override registry, and that registry ships a hand-crafted derivative template. Not worth the maintenance for a single-call-only experience when Qwen/Mistral/Llama/Hermes are natively, robustly tool-capable. glm-4-9b now reports a clean 'does not support tool calls' like the other unsupported families. The override mechanism was never merged, so nothing to remove.

PR #391, opened 2026-06-19
Per-architecture chat-template overrides for tool-aware fleet serving

tobocop2:
## Problem

The fleet serves chat with `llama-server --jinja`, which renders the GGUF's embedded chat template and parses native tool calls from it. Some published quants ship a template with no tools block, so the server cannot emit tool calls and lilbee declines to offer tools at all (`supports_tools` is False) even for models that can tool-call.

## Solution

Add a per-architecture chat-template override. For an architecture whose embedded template strips tools, vendor the model's tool-aware upstream template and serve it via `--chat-template-file`, but only when the embedded template is not already tool-aware, so a quant that ships a good template (e.g. arch `llama` for Llama 3.1) is never touched. One resolver feeds both seams, the launch argv and the `supports_tools` decision, so a model is offered tools exactly when it is served a tool-aware template.

The registry is seeded with `lfm2` only, verified on a real fleet to now emit a structured, dispatchable tool call where it previously fail-fasted with 'does not support tool calls'. Other candidate archs were deferred after the same fleet run showed their tool-call output is not parsed end-to-end: the override makes the model offer tools, but it emits the call as prose that `--jinja` does not turn into a structured call, so the agent never dispatches and the model hallucinates. Offering tools we cannot parse is worse than the honest fail-fast, so those archs are left out until llama.cpp parses their tool-call output.

Tests cover the resolver, both seams, and a render of each vendored template on the no-tools warmup payload (a guard against a template crashing the server's jinja engine on load). Verified end-to-end on a single-GPU fleet.


tobocop2:
Closing without merging. The override registry only ever had one viable entry, lfm2 (LFM2-1.2B-Tool), and the fleet verify showed the other candidate archs cannot be fixed by a template (their tool-call output is not parsed by llama.cpp, or they have no tool-aware upstream template). lfm2 itself is a niche model (~500 downloads/month) and fails multi-turn anyway. The popular tool-callers already ship tool-aware templates and need none of this. Not worth carrying the hook, vendored template, and packaging to serve one low-adoption model, so we are dropping it rather than maintaining it.

PR #392, opened 2026-06-19
Detect NVIDIA GPUs without initializing NVML in-process

tobocop2:
## What this does

`has_nvidia_gpu()` / `_try_nvidia_memory()` could call `pynvml.nvmlInit()` **in the main process**. Initializing NVML (or CUDA) in the process that later spawns the `llama-server --list-devices` probe risks leaving driver state that makes the probe enumerate no device — a latent foot-gun on hosts where `pynvml` is installed.

This makes GPU detection run **only as subprocesses** — `nvidia-smi` first (ships with the driver), then an isolated `python -c` + `pynvml` fallback — and caches the result. With no in-process NVML/CUDA init, GPU detection can never poison the device probe.

## Honest scope

This is a **defensive robustness fix**, not the root cause of the Phase-2 ingest failures we hit (on those pods `pynvml` wasn't installed, so detection already used the nvidia-smi subprocess; the actual fix there was #389's `apply_cuda_runtime_env`, now merged). It removes a real latent failure mode for setups that do have `pynvml`, and keeps GPU detection from ever interfering with the engine probe.

Unit tests cover nvidia-smi primary, the isolated-pynvml fallback, and an explicit assertion that NVML is never imported in-process. Full lint + typecheck green.

### DIFF
--- a/src/lilbee/providers/model_cache.py
+++ b/src/lilbee/providers/model_cache.py
@@ -2,8 +2,11 @@
 
 from __future__ import annotations
 
+import functools
 import logging
 import platform
+import subprocess
+import sys
 from enum import StrEnum
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


tobocop2:
Closing: this is a defensive robustness change (GPU detection never initializing NVML in-process), but it is NOT the root cause of the Phase-2 fleet-engine failures, so it should not be carried as 'the fix.' The real issue (the vision/embed servers not coming up / producing empty output / GPU idle in the sync pipeline) is still un-root-caused and is being investigated properly. If the NVML isolation is wanted later as a standalone hardening, it can be reopened on its own merits.

PR #429, opened 2026-06-22
Configurable model placement

tobocop2:
## Problem

lilbee places models across GPUs automatically with a VRAM-aware planner, but there is no way to override that decision or even see it. On a heterogeneous box, a card reserved for something else, or a model the heuristic splits badly, placement is a black box you cannot steer.

## Solution

Adds a manual placement spec. Per role you pin the GPU devices, and optionally set the tensor-split ratio and replica count. When a spec is present it replaces the automatic planner; a spec that does not fit VRAM fails fast and names the card that is short.

Placement is now visible and editable from four surfaces, all over one shared use-case: the CLI (`lilbee placement show/preview/set/clear`), the HTTP API, MCP, and a TUI screen. Preview is a dry-run that shows what auto placement (or a candidate spec) would choose, including each card's identifier and free VRAM, so you can check a change before applying it. The spec persists in config.toml and survives a restart.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -53,6 +53,22 @@ When reporting a worker crash, attach the last ~100 lines of the worker log plus
 tail -n 100 "<data root>/logs/worker-embed.log"
 ```
 
+## Fleet refuses to start after a hardware change
+
+If you have a manual placement spec set and a GPU is removed, replaced, or
+renumbered, the fleet will refuse to start with an error naming the card that
+no longer fits. The error is intentional: lilbee won't silently start in a
+broken state.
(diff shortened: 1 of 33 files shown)


tobocop2:
Review note: the generic settings-write path (e.g. setting `placement` via `lilbee set`/the settings API) does not run the VRAM fit-check that `lilbee placement set` performs, matching the existing behavior of sibling fields like `gpu_devices`/`n_gpu_layers`. Use the placement surfaces (`lilbee placement set`, the TUI screen, MCP/HTTP) to get the fit validation + fleet reload.

tobocop2:

PR #433, opened 2026-06-22
Migrate document extraction to xberg 1.0

tobocop2:
## Problem

Document extraction ran on kreuzberg 4.x, which split text, OCR, chunking, table structure and layout into separate passes and leaned on its own bundled defaults for OCR, chunk boundaries and token sizing. The extraction code was also scattered across loose modules under `data/` and `data/ingest/`.

## Solution

Migrate to xberg 1.0.4, whose single async `extract()` owns text, OCR, chunking, tables and layout in one pass. lilbee registers its own vision model, embedder and tokenizer as xberg plugin backends, so scanned-page OCR, chunk-boundary detection and token-budgeted sizing route through the fleet instead of xberg's defaults. Vision OCR and PDF reading go through xberg directly, replacing the hand-rolled page rasterization.

Ingest now returns `(records, SourceMeta)`, so the document title and authors reach the source row and every chunk (prefixed to embeddings when `embed_titles` is on). Table extraction and layout-aware PDF reading are opt-in; force-OCR drives `ExtractionConfig.force_ocr`.

## New extract package

The extraction subsystem moves into a `data/extract/` package (mirroring `data/store/`): the xberg bridge, the document/markdown drivers, batching, chunking and the tree-sitter code chunker, plus a `backends/` subpackage for the OCR/embedding/tokenizer plugins. Shared primitives (`offload`, `title`, `types`) sit at the `data/` root, so `store`, `extract` and `ingest` form a clean layered graph with no import cycle. `data/ingest/` stays the corpus-orchestration layer.

## Improvements

- Extraction covers 100+ formats, tracked from xberg's own list: Office including legacy `.doc/.xls/.ppt` plus `.odt/.rtf`, ebooks, email, archives, and more image types.
- OCR runs in 100+ languages through Tesseract (`LILBEE_OCR_LANGUAGE`, e.g. `eng+deu`), alongside table- and layout-preserving vision OCR.
- Intel Mac binaries return (native `macos-15-intel` build) and the macOS floor drops, since xberg supports a lower baseline.
- Supported model architectures are read from the pinned engine's own arch table rather than a floating package.


src/lilbee/app/services.py
```diff
@@ -256,6 +269,62 @@ def services_scope(services: Services) -> Iterator[None]:
...
+    """Register lilbee's embedder as xberg's plugin embedding backend.
+
+    The semantic chunker uses it to detect topic boundaries with the same model
+    that vectorizes chunks, instead of xberg's bundled ONNX preset. The backend
+    reads ``cfg.embedding_dim`` live and routes through ``provider.embed`` (which
+    resolves the embedding model per call), so a model swap needs no
+    re-registration -- but it captures ``provider.embed``, so it must re-bind
+    whenever the provider is rebuilt (``reset_services``).
+    """
+    from xberg import (
```
Goldziher: in general its best to do top level imports, unless you have a good reason to do them lazily. Python is not very efficient in this - loading modules on demand is expansive. Obviously it depends on cold start speeds - and what is important for you. But I see this is a pattern you use commonly, so I would actually recommend you move away from it - except where you need lazy loading (note lazy is a new keyword in python 3.15, which will resolve this issue)

tobocop2: So I agree and I definitely have unnecessary lazy imports in the codebase that I'm going to address shortly. This one I can justify, as lilbee has a cli and chat that don't touch extraction. keeping the import lazy allows me to keep startup and adjacent things as fast as possible until extraction becomes necessary. I'm now questioning the design so I appreciate the call out and want a cleaner solution so i'm going to explore and report back.

PR #434, opened 2026-06-23
Fleet: elastic ingest pool, VRAM-aware replica scale-down

tobocop2:
## Problem

The multi-GPU fleet sizes embed and vision replicas for ingest throughput and then keeps them resident in VRAM forever (the generated llama-swap config sets `ttl: 0` on every member, with no idle-unload). After an ingest, the embed replicas stay pinned. When a large chat model loads on top, VRAM saturates, llama-swap thrashes and re-listens on new ports, and lilbee loses the chat server handle, surfacing "No chat model server is running" even though a chat process is alive (bb-naa). Replica counts also default to 1 and never use the extra GPUs, while a naive auto-default to GPU count would make the crowding worse (bb-s61).

## Solution

The fleet now reserves a persistent query fleet first (chat, one embed, rerank, one vision) and places the extra ingest replicas only into the VRAM that remains, so a chat issued during an ingest always fits. When an ingest finishes, the extra replicas are unloaded one by one via llama-swap's `/api/models/unload`, freeing their VRAM without disturbing the resident chat (the unload runs off the request-serving loop). `embed_replicas` and `vision_replicas` default to one replica per GPU, capped by residual VRAM, and the 0-means-auto value is resolved the same way everywhere it is read. A dead or restarted llama-swap is now re-probed and rebuilt once before the fleet reports no server.

100% coverage, mypy strict, full suite green.

Validation: the VRAM-reclaim behavior of `/api/models/unload` is being validated on a multi-GPU pod. A two-swap fallback is documented if that endpoint does not cleanly reclaim VRAM.

Note: the placement-interaction guard (skip the pool release when a manual placement spec is set) is deferred until the manual-placement feature (#429) merges, since `cfg.placement` is not on this base branch.

Closes bb-naa, bb-s61.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -62,6 +62,22 @@ tail -n 100 "<data root>/logs/llama-swap.log"
 tail -n 100 "<data root>/logs/server.log"
 ```
 
+## Ingest replicas and VRAM reclaim
+
+When an ingest job finishes, any extra embed or vision replicas that were started
+for it are unloaded automatically. This frees their VRAM so chat and search
+capacity is restored without a restart.
+
+If you set `embed_replicas` or `vision_replicas` explicitly in your config, that
+count is used during ingest and reclaimed the same way afterward. The default
+(`0`) picks one replica per GPU, capped by whatever VRAM remains after the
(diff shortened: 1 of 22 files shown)


tobocop2:

PR #437, opened 2026-06-24
Placement QA fixes: usable TUI editor + auto-split for large chat models

tobocop2:
## Problem

Two issues surfaced while QA-ing the placement feature on real multi-GPU hardware:

1. **The TUI placement editor ignored everything typed into it.** The screen parsed a private field that was never updated from the spec TextArea, so pressing Apply or Preview acted on an empty spec; the auto placement re-resolving to different cards made it *look* like the edit had applied. Every existing test masked this by assigning the field directly instead of driving the widget. Beyond the bug, a raw-JSON text box is a poor primary UX for picking GPUs per role.

2. **Auto placement collapsed a large chat model's context to ~512 tokens.** A ~17GB model fits one 24GB card under the planner's 0.9 headroom, so it was single-carded; but the served KV cache is sized at `gpu_memory_fraction` (0.75), leaving almost no room, so `n_ctx` dropped to ~512 and real chat/agent prompts returned 400. The user had to hand-write a placement spec to split the model across cards just to make it usable.

## Solution

**TUI:** read the live editor instead of a stale field, and replace the JSON box with an interactive per-role editor: GPU toggle buttons and a replica stepper per role, a live auto/manual/edited title, a Roles column on the GPU table, and a read-only equivalent-spec line (so the JSON is still copyable for the CLI/HTTP/MCP surfaces). Action bindings are priority so they fire from the editor. Tests now click the real widgets.

**Planner:** charge the chat placement footprint against the serve budget rather than the placement headroom (and reserve KV for the ctx target, floored at a usable minimum). A model that would starve its context on a single card is now tensor-split to pool VRAM across cards automatically. Small models are unaffected.

Validated on a 2x RTX A5000 box: auto placement now splits a 30B coder model (`devices=[1, 0]`) and a query succeeds with no manual spec, where before it errored on the 512-token window. Fixes bb-9rn.

### DIFF
--- a/docs/placement_ux_prototype.py
+++ b/docs/placement_ux_prototype.py
@@ -1,0 +1,446 @@
+"""Placement screen UX prototype — calm overview + per-model drill-in + advanced controls.
+
+A design reference for the real TUI placement screen, not production code.
+Grounded in lilbee's real fleet: 4 models (Chat, Embeddings, Reranker, Vision).
+Chat/Reranker are single instances (tensor-split a big model across cards);
+Embeddings/Vision are data-parallel (a copy per card) for throughput. LM-Studio
+style: an overview, then one model at a time. Fake data, no fleet.
+
+The drill-in has an Advanced section (press 'a') that exposes the full
+RolePlacement surface: distribute mode (copies vs split) for replicable roles,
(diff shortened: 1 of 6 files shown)


tobocop2:

PR #452, opened 2026-06-29
Fix placement back-navigation crash

tobocop2:
## Problem
The placement editor's back action (`action_go_back`) used a raw `pop_screen()`. Placement is entered as a managed view via `switch_view` (a `switch_screen` *replace*, not a `push_screen`), so leaving it through `pop_screen` underflowed Textual's result-callback stack on a rapid second back, crashing the TUI with `IndexError: pop from empty list` (textual/screen.py `_pop_result_callback`). Intermittent — a timing race, easy to hit under rapid navigation.

## Solution
Route the back action through the guarded `switch_view("Chat")`, the inverse of how Placement is entered, so a racing second transition is dropped by the existing `_switching` guard instead of underflowing the screen stack. The now-dead `stack > 1` / `pop_screen` branch is removed.

Regression tests drive rapid back-to-back placement transitions and assert no underflow (the test reproduces the exact `IndexError` on the old code and passes on the fix), plus an escape-returns-to-Chat test. ruff/format/mypy-strict clean; TUI navigation + placement suites green.

### DIFF
--- a/src/lilbee/cli/tui/screens/placement.py
+++ b/src/lilbee/cli/tui/screens/placement.py
@@ -308,8 +308,5 @@ class PlacementScreen(Screen[None]):
             call_from_thread(self, setattr, self, "applying", False)
 
     def action_go_back(self) -> None:
-        """Pop back to the previous screen."""
-        if len(self.app.screen_stack) > 1:
-            self.app.pop_screen()
-        else:
-            self.app.switch_view("Chat")
+        """Return to Chat via the guarded switch_view, the inverse of how Placement is entered."""
+        self.app.switch_view("Chat")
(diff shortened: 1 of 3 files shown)


tobocop2:
Fix is already incorporated in #267, adapted to the renamed Fleet view. This branch edits the old screens/placement.py + tests/test_tui_placement.py, but feat/local-model-api renamed those to screens/fleet.py + tests/test_tui_fleet.py. Merging as-is would resurrect the deleted placement.py and conflict. The bb-ce4 back-nav fix lives in fleet.py (action_go_back routes through the guarded switch_view) with the rapid-back regression test ported to test_tui_navigation.py (test_rapid_fleet_back_does_not_corrupt_screen_stack). Closing as folded-in rather than merging a stale-path branch.

PR #453, opened 2026-06-29
Native GPU fleet panel in the TUI (placement screen)

tobocop2:
## Problem
The live per-GPU "GPU FLEET" view (utilization + VRAM bars) only existed as an external script in a tmux split for demos — there was no native in-TUI GPU monitor. The placement screen showed a static GPU table but no live load, so a faithful demo of the multi-GPU fleet meant compositing a separate process beside the TUI.

## Solution
Add a `GpuFleetPanel` widget on the Placement screen that renders per-card utilization + VRAM bars (rose-pine) and refreshes on a ~1s tick. It reuses the existing `gpu_stats.probe_gpu_stats` sampler, run off the UI thread (`@work(thread=True)` + `call_from_thread`) so the event loop is never blocked by the probe subprocess. VRAM is cross-vendor (from the device probe); utilization is NVIDIA-only today — the per-vendor backends (AMD/Apple/Intel) + temperature are the follow-up (bb-dzc). Graceful empty state when no GPUs or `nvidia-smi` is absent, and a transient probe failure keeps the last content rather than blanking.

Tests drive the widget and the placement integration via pilot (empty state, util/VRAM rendering, the None-utilization dash, tick updates, probe-failure resilience, timer cleanup on unmount). CSS lives in a `.tcss` loaded via the widget `DEFAULT_CSS` convention; ruff/format/mypy-strict clean.

### DIFF
--- a/src/lilbee/cli/tui/screens/placement.py
+++ b/src/lilbee/cli/tui/screens/placement.py
@@ -24,6 +24,7 @@ from textual.widgets import Button, DataTable, Footer, Label, Static
 from lilbee.app.placement import get_placement, preview_placement, set_placement
 from lilbee.cli.tui.app import LilbeeApp
 from lilbee.cli.tui.thread_safe import call_from_thread
+from lilbee.cli.tui.widgets.gpu_fleet_panel import GpuFleetPanel
 from lilbee.providers.fleet.placement_spec import PlacementError, PlacementSpec, RolePlacement
 from lilbee.providers.roles import WorkerRole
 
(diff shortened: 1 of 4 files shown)


tobocop2:
Subsumed into #267: the Fleet view (squashed, hardcoded-theme commit removed) was consolidated into feat/local-model-api along with the cross-vendor gpu-stats backends and OS-hardening. Content is tree-identical to this branch's HEAD. Closing as folded-in.

PR #467, opened 2026-07-01
build: native Intel-Mac standalone binaries (macos-15-intel)

tobocop2:
## Problem

lilbee shipped no download-and-run standalone binary for Intel Macs. The
`pip` path already works there (main cross-compiles `macosx_11_0_x86_64`
wheels and documents `--extra-index-url https://lilbee.sh/cpu/`), but the
standalone executable was dropped back when the old `macos-13` Intel runner
was flaky, leaving Homebrew/direct-download users on Apple Silicon or Linux
only.

## Solution

GitHub now offers a native Intel image, `macos-15-intel` (the old `macos-13`
is retired). Nuitka can't cross-compile arch from the arm64 runner, so this
builds natively there and runs the full smoke test on real Intel hardware —
no cross-compiled-and-unverified asset.

Two soft cells, mirroring the existing linux/windows pattern:

- `lilbee-macos-x86_64` — stock AVX baseline. Covers every Mac that can run
  macOS 11+ (all Ivy Bridge or newer); deployment target stays 11.0.
- `lilbee-compat-macos-x86_64` — pre-Haswell build (launcher held to
  `x86-64-v2`, stock lancedb swapped for `+compat`), covering the
  AVX-but-no-AVX2 tail such as the 2013 Mac Pro and guarding against an
  AVX2-baselined lancedb SIGILL.

Both are `continue-on-error`, so a queued or missing Intel runner never
blocks the release fan-out. A pre-AVX (2008-era) engine was considered and
skipped: no Mac that runs macOS 11/12 is pre-AVX, so it would help nobody
without also lowering the OS floor to High Sierra.

README lists the new compat binary in the older-CPU download row.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -157,7 +157,25 @@ jobs:
           - os: macos-latest
             asset_name: lilbee-macos-arm64
             backend: metal
-          # Intel Mac dropped: macos-13 hosted runners stay queued; pip wheel covers them.
+          # Intel Mac: built natively on macos-15-intel, GitHub's Intel image
+          # (Nuitka can't cross-compile arch from the arm64 runner). Soft because
+          # the Intel image can queue longer than the arm64 pool -- this asset must
+          # never block the release fan-out. Deployment target stays 11.0, so the
+          # AVX baseline covers every Mac that can run macOS 11+ (Ivy Bridge or newer).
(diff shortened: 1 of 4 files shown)


tobocop2:
will need to revisit later and add the same support to xberg

PR #472, opened 2026-07-03
True-speed multi-GPU demos: embed and chat in both placement modes

tobocop2:
## Problem

The placement and chat demos on gh-pages predate the per-role fleet work and the kreuzberg-5 integration. The old chat reel asked a trivial oil-capacity question, and the placement reels showed the previous editor UI.

## Solution

Re-recorded all three reels on a 3x H100 pod running the trial build that merges feat/kreuzberg-5 (which includes the local-model-api work): Qwen3 235B thinking chat split across two cards, three Qwen3 8B embedding replicas.

- **tui-multi-gpu-self-index.gif**: live `/add` of the source tree with the placement drawer open while all three GPUs index; the resident 235B stays loaded throughout, then answers a question about its own placement code.
- **tui-manual-placement.gif / placement-manual.{gif,mp4,png}**: manual placement editing in the drawer with a live apply; the embedding pool consolidates from three replicas to one in a couple of seconds while the chat model never reloads.
- **tui-chat.{gif,mp4,png}**: ask the Crown Victoria service manual. Towing pre-trip check where the model catches that a 3,500 lb trailer exceeds the manual's 1,500 lb limit, then a blower/AC fuse diagnosis with cited fuse numbers and panel locations. The gif is the fuse-diagnosis excerpt sized for README embedding; the mp4 carries both questions.


tobocop2:
Superseded by a fresh demos branch cut from latest main with the search overhaul folded in. The reels here predate a lot of TUI and retrieval changes, so they're being re-recorded from scratch rather than patched. A new PR will replace this.

Issue #474, opened 2026-07-03
Fleet rebuild after a placement change takes over a minute


tobocop2:
Applying or clearing a GPU placement rebuilds the whole fleet through the cold-boot path, and on a serious box that takes over a minute. From the plugin or the TUI it reads as the app hanging.

## The symptom

On a 3x A100 box serving Qwen3-235B Q3_K_M (112GB, 3 shards) plus Qwen3-Embedding-8B, any placement change (`set_placement` -> `reset_services` -> `warm_up_pool`) takes 72 to 88 seconds before `chat_ready` flips back to true. Even moving just the small embedder pays the full price, because the whole fleet is torn down and rebuilt from scratch.

## Where the time goes

Measured from llama-swap's request log plus `nvidia-smi` sampled at 1.5s intervals during rebuilds:

| Phase | Time | What it is, and why it is waste |
|---|---|---|
| llama-swap full restart | ~6s | Every placement change tears down and respawns the whole proxy, even when one role moved. |
| Light roles warm first | ~7s | `_preload_roles` warms roles in a sequential for loop and chat happens to come last, so the big model waits behind the small ones. |
| Shard preread | ~15-20s | `_prewarm_chat_weights` rereads all 112GB of shards on every rebuild. The read exists to warm the page cache for cold network-filesystem loads, but on a rebuild those pages are already hot, so it is pure waste. |
| Chat server bring-up | ~40s | The VRAM allocation appears in ~3s; the rest is weight upload plus init at roughly 3.7GB/s aggregate, then KV allocation right before ready. |

```mermaid
flowchart LR
    change["placement change"] -->|"~6s"| restart["llama-swap<br/>full restart"]
    restart -->|"~7s"| light["embed + rerank warm<br/>(sequential, chat last)"]
    light -->|"~15-20s"| preread["reread 112GB of shards<br/>(already in page cache)"]
    preread -->|"~40s"| bringup["chat llama-server bring-up<br/>weight upload + KV alloc"]
    bringup --> ready["chat_ready = true<br/>72 to 88s total"]
```

## Quick wins, already verified on the box

Two orchestration fixes cut rebuilds to 47 to 49 seconds; PR to follow:

- Warm the roles concurrently, so the light roles hide behind the chat load instead of running before it.
- Skip the shard preread when those shards were already paged in this boot (module-level marker keyed on path, size, and mtime).

## What still needs investigation

The remaining ~40s is the engine bring-up itself:

[...]

PR #476, opened 2026-07-03
Cut fleet rebuild time by warming roles concurrently and skipping redundant shard rereads

tobocop2:
Applying or clearing a GPU placement reruns the cold boot warm path, and on a 3x A100 box serving Qwen3-235B Q3_K_M it took 72 to 88 seconds before chat came back. Profiling (issue #474) showed most of that time never touches the model load itself: `_preload_roles` warmed roles in a sequential loop so the chat load queued behind the embedder and reranker, and `_prewarm_chat_weights` reread all 112GB of shards on every rebuild even though the page cache was already hot from the previous load.

Two changes, both in the warm path. Roles now warm concurrently, which matches the co-resident swap group (`swap: false, exclusive: false`) and the architecture doc's "brings every role up at once" description; the light roles load alongside the chat long pole instead of before it. And the shard preread now records what it has paged this boot (keyed on path, size, and mtime) and skips the reread when nothing changed, so it only pays the read once per boot per model. A replaced or redownloaded shard gets a new key and rereads as before.

Measured on the A100 box: rebuilds drop from 72-88s to 47-49s. The remaining time is llama-server's own engine bring-up, tracked separately in #474.

### DIFF
--- a/src/lilbee/providers/fleet/provider.py
+++ b/src/lilbee/providers/fleet/provider.py
@@ -68,6 +68,10 @@ _WARM_MAX_TOKENS = 1
 # Read size for paging chat shards into the page cache during warm; large enough
 # to keep sequential reads efficient without holding much resident at once.
 _PREWARM_CHUNK_BYTES = 8 * 1024 * 1024
+# Shards fully paged in this boot, keyed on (path, size, mtime_ns); a fleet
+# rebuild (e.g. a placement change) skips re-reading a hot cache. Module-level so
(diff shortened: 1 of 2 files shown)


tobocop2:
Superseded by #477, which carries this branch's commit unchanged (concurrent role warms + prewarm skip) and adds the per-role llama-swap split so placement changes stop restarting untouched roles, plus the measured --no-mmap chat load. Closing in its favor.

PR #480, opened 2026-07-04
Ingest trace logging + three extraction-pipeline fixes found on a production run

tobocop2:
## Problem

A large batch-ingest run surfaced three silent-loss bugs in the extraction pipeline, and diagnosing them required per-file visibility the pipeline didn't have.

1. OCR request tokens vanished across the xberg native boundary: `OcrConfig.backend_options` comes back as a JSON string, not the dict lilbee put in, so the per-page progress callback and the configured `ocr_timeout` were silently dropped on every vision OCR call.
2. Embed inputs were tail-truncated: the embed server context was sized in tokens from `chunk_size`, but the chunker enforces a character budget (`chunk_size * CHARS_PER_TOKEN`). Token-dense text (numeric tables, dense identifiers) tokenizes near 1.5-2.5 chars/token, so real chunks approached 2000 tokens against a 512-token cap and lost their tails — silently unsearchable text.
3. Trace lines never reached disk under the TUI, whose root file handler is WARNING+.

## Solution

- Per-file extraction tracing (`LILBEE_INGEST_TRACE=1`): one parseable line per file (source, type, elapsed, pages, chunks, ocr_pages, vision attribution) on dedicated loggers, plus a vision line when OCR fires. `LILBEE_INGEST_TRACE_FILE` attaches a file handler so the destination is independent of the host front-end.
- The OCR config view accepts both dict and JSON-string `backend_options`, restoring progress ticks and timeouts.
- Embed context is now the chunker's character budget capped by the embedder's trained context. A BPE token is at least one character, so that bound is provable: embed-time truncation cannot occur for chunker-emitted chunks. Verified on a live multi-GPU run — zero truncation warnings end to end, where the previous heuristic caps truncated every long chunk (1x) or ~2% of chunks (2x).

All three fixes are exercised by unit tests; the trace format is covered including the WARNING-root and JSON-string regression cases.


tobocop2:
Superseded by a clean PR carrying only the pipeline fixes.

PR #520, opened 2026-07-11
Fix vision OCR refused on consumer GPUs: correct the projector estimate and load tight models best-effort

tobocop2:
## Problem

Ingest OCR still fails on consumer GPUs after the last placement fix: lilbee refuses to serve the vision model with "does not fit available memory", so scanned documents come back empty. Two defects stack.

### The vision memory estimate is wrong

The memory estimator overstates any vision model's GPU footprint by about 10 GiB. A 1.2 GB OCR model is estimated at 12.7 GiB, larger than most consumer cards, so vision can never pass the fit check no matter how small the model really is.

Root-causing this exposed a bug in gguf-parser itself: its discrete-GPU accounting for multimodal projectors of unrecognized types skips the spatial-merge reduction and sizes a quadratic attention buffer against the full-resolution patch grid. Reported upstream with a reproducer and analysis: gpustack/gguf-parser-go#21.

### One bad estimate kills the role

Placement treats the estimate as a hard gate: a model that does not fit by estimate gets no server at all. The pre-fleet engine never refused a load. It tried, and the load itself decided.

### Work around the estimator bug

Until gguf-parser fixes the projector accounting upstream, this PR carries a workaround: the vision estimate is rebuilt from the accounting the estimator gets right (its unified-memory figures), with the projector's own file size as a floor. The same OCR model now estimates 3.8 GiB, matching its real usage. Once the upstream fix ships and the pin is bumped, the workaround can be re-evaluated; it stays accurate either way.

### Warn instead of refuse

On GPUs the estimate now only advises. A model that does not fit by estimate still gets a server: it is placed on the card with the most free memory and lilbee logs a warning that memory is tight and the load may be slow or fail. If the load truly fails, the existing per-page OCR skip handling applies, so the worst case is no worse than today's refusal. This also means any future estimator error degrades to a warning instead of a dead role.

CPU and Apple Silicon hosts keep the strict check, since an oversize load on shared memory can freeze the whole machine. Manually pinned layouts are also still validated strictly.

### No change on larger machines

The warn-instead-of-refuse path only triggers where today's code refuses outright. A regression test proves the warning can never fire for a placement the memory budget affords.

## Testing

The oracle parity harness now asserts the GPU planner never refuses a role, across 14 low-memory and on-the-cusp scenarios including this exact failure. The full suite is green at 100% coverage, and the estimate fix is verified against the real gguf-parser and a real vision model.

Verified end to end on real hardware: a rented 12 GB RTX 3080 Ti ran the same scanned-PDF ingest and query on v0.6.66b507 (the pre-fleet baseline), the current release, and this branch. The current release reproduces the refusal verbatim; this branch serves every scenario the baseline served, including a forced memory-is-tight case where the warned load succeeds and OCR completes.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -305,7 +305,12 @@ flowchart LR
   whichever GPU is fullest, not on the summed total. The estimate also carries the
   same `--batch-size`/`--ubatch-size` the launch will use (pooled embed/rerank raise
   them to the full context), so the compute-buffer estimate matches what the server
-  actually allocates. This replaced a hand-rolled
+  actually allocates. For a vision model the discrete-GPU number is corrected on
+  lilbee's side: gguf-parser's `nonuma` merge charges any multimodal projector
+  roughly 10 GiB of phantom compute buffer (a 1.2 GiB OCR model plus its 0.9 GiB
+  projector came back as 12.7 GiB), so the projector is instead charged at its
+  unified-memory delta, floored at the projector's own weights. This replaced a
+  hand-rolled
   weights + KV-cache estimate that used discrete-GPU accounting and over-estimated
   ~3x on unified memory, which was crowding the co-resident embed/rerank servers out
   of the budget and 503-ing every search.
@@ -451,8 +456,13 @@ touching the running fleet.
   `swap: true` group evicts same-group siblings, so a second replica inside it would
   evict the first). Interleaving a query with an ingest costs a model reload each
   way. With the VRAM to hold everything at once, no swap group forms and every role
-  pins to its own group. A role is unplaceable only when it does not fit even beside
-  the one embedder its own phase needs (a genuine "use a smaller model").
+  pins to its own group. On GPUs the estimate advises but never refuses: a role that
+  does not fit even beside the one embedder its own phase needs is still **placed
+  tight** (best-effort) on the emptiest card, anchoring the swap group so the load
+  starts against as much free VRAM as possible, with a memory-is-tight warning
+  carrying the estimated shortfall. The in-process pool never gated a load on an
+  estimate, and the load itself is the only reliable arbiter (the estimate can be
+  high, and some drivers page into shared memory rather than fail).
 - **Data-parallel replicas** (`embed_replicas` / `vision_replicas`): the embed and
   vision roles can run as N independent servers, fanning embedding and OCR across
   the box during ingest. Setting either value to `0` (the default) means auto: one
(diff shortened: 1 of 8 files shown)


tobocop2:
## Hardware model-load sweep results

Verification harness in the style of the gguf-parser estimate sweep, applied at the load layer: 23 unique GGUF files (8 vision families covering every distinct multimodal projector code path, plus chat/embed/rerank models) each driven through the real pipeline on a rented 12 GB RTX 3080 Ti: estimate, placement, llama-swap launch, actual inference. Fix branch vs the current release (dev716) on identical hardware, config, and an image-only scanned PDF whose content is reachable only through vision OCR. The probe question's answer (42 newton meters) exists only in the scanned pixels.

### Vision families

| Family (projector type) | dev716 | fix: refused | fix: OCR | fix: answer has content | placement estimate |
|---|---|---|---|---|---|
| LightOnOCR-2-1B (lightonocr) | **refused: the shipped bug** | no | worked | **yes** | 3.76 GiB |
| Qwen2.5-VL-3B (qwen2.5vl merger) | served | no | worked | **yes** | 3.81 GiB |
| Qwen3-VL-4B (qwen3vl) | served | no | worked | **yes** | 7.72 GiB |
| gemma-3-4b-it (gemma3) | served | no | worked | **yes** | 4.71 GiB |
| SmolVLM2-2.2B (idefics3) | served | no | worked | **yes** | 2.94 GiB |
| MiniCPM-V-2.6 (resampler) | served | no | worked | **yes** | 7.33 GiB |
| InternVL3-2B (internvl) | served | no | worked | no (2B answer quality) | 2.68 GiB |
| LLaVA-1.6-7B (mlp) | served | no | worked | no (older model quality) | 9.20 GiB |

The dev716 refusal hits exactly the projector types gguf-parser's estimator does not recognize (gpustack/gguf-parser-go#21): LightOnOCR here, and the same class in the original user report. Known projector types were never refused, which matches the root cause. The two content-negative rows loaded and OCR'd cleanly; the answers were wrong for model-capability reasons, reported rather than hidden.

### Role models (fix branch)

Llama-3.2-3B chat, Qwen3-Embedding-0.6B, and bge-reranker-v2-m3 all loaded and answered (rc 0 throughout). The embed scenario surfaced a pre-existing transient chat-load OOM that self-recovered on retry; filed separately, identical on dev716, unrelated to this PR.

### Invariants

- fix refusals: **0 of 8** (must be 0)
- fix OCR-backend failures: **0 of 8** (must be 0)
- fix vision loads: **8 of 8**
- Peak VRAM per scenario: ~10.9 GB on the 12 GB card, at the 90% budget line by design

### Notes

- Five families were initially blocked on both builds identically by a pre-existing, unrelated defect this sweep discovered: several mainstream VL repos are classified as chat models, so their multimodal projector is never downloaded and planning silently drops the vision role. Filed separately with a repro; the sweep worked around it by placing the projector files directly, after which all five families passed as shown above.
- Earlier hardware rounds on the same card class: the pre-fleet in-process build (v0.6.66b507) as oracle baseline, dev716 reproducing the user-reported refusal verbatim, and a forced memory-is-tight scenario where the warned load succeeds and OCR completes. Full write-up: https://gist.github.com/tobocop2/80293ae854b0c58b0974558043a13af9

PR #531, opened 2026-07-13
Release channels: build once, promote dev to beta to stable without rebuilding

tobocop2:
## Problem

Promoting a build to a different maturity level (dev to beta, beta to stable) meant cutting a new tag and rebuilding every artifact, even when nothing changed but the label. The version string was the only thing distinguishing a dev build from a beta from a stable release, and since it is baked into the wheel metadata, the binary's `--version`/`/api/health`, and PyPI's immutable version index, any relabel forced a full pipeline run. Separately, every tag overwrote the single stable manifest of every package channel, so dev builds shipped straight to everyone's default `brew upgrade`/`scoop update`, and docker `:latest` moved on every dev tag.

## Solution

Builds happen exactly once per tag; a channel (dev/beta/stable) is a pointer to a tag, recorded in `packaging/channels.json` on main. The new `promote.yml` is the single promotion door and only moves pointers and republishes the tag's existing GH-release assets:

| Channel | Entered by | Gate | Publishes |
|---|---|---|---|
| dev | automatic, every release candidate | build + artifact smoke | docker `:dev`/`:dev-cuda` |
| beta | `make promote TAG=... CHANNEL=beta` | green `verify-release` run | docker `:beta`, homebrew/scoop/AUR/nix `lilbee-beta` variants |
| stable | `make promote TAG=... CHANNEL=stable` | green `verify-release` run + prior beta soak | PyPI, docker `:latest`/`:cuda`, unsuffixed packages + flatpak, CUDA packages, GH release flipped to latest with headed notes |

Rollback is promoting an older tag again. A `force=skip-verify` input replaces the old `emergency-publish.yml` skip-QA lane, and the old `fanout-packaging` action is absorbed into promote's dispatches.

Supporting changes that make this work end to end:

- **Final-form versions.** `release.sh` now cuts `0.6.90`, `0.6.91`, ... instead of `bN.devN` counters, since pip's default resolution would forever skip a stable promotion of a marker-carrying version (the stable gate refuses them). The current `0.6.90b420.dev719` graduates to `0.6.90` on the next cut. GH releases are always created as prereleases; stable promotion flips the flag (prerelease-ness comes from the channel now, not the version string).
- **PyPI is the stable channel.** `publish.yml` runs at stable promotion and downloads the wheel + sdist from the GH release rather than from run artifacts, so promotion still works after the 90-day artifact retention. The sdist is now attached to the release alongside the wheel. The trusted-publisher pins (`publish.yml` filename, `pypi` environment) are untouched.
- **Beta package variants.** Seeds for `Formula/lilbee-beta.rb`, `bucket/lilbee-beta.json`, `packaging/aur/lilbee-beta`, and `sources-beta.json` + a `lilbee-beta` flake output, following the existing lilbee-compat variant pattern; all renderers were dry-run against the new seeds. The homebrew tap formula self-seeds on first publish, and the AUR `lilbee-beta` package is created by the first beta promotion. Flatpak stays stable-only for now (its OSTree repo would need a branch per ring); beta users there use docker `:beta` or the GH release binary.

Behavioral changes to be aware of: dev builds no longer touch PyPI, package manifests, or docker `:latest`; those surfaces only move at explicit promotions. `make release-promote` is replaced by `make promote TAG=... CHANNEL=...`. Docs updated in `docs/architecture.md` (release pipeline section) and `docs/development.md`.

All workflows pass actionlint; JSON/ruby/shell seeds syntax-checked; the flake change is covered by `flake-check.yml` on this PR (trigger paths now include `sources-beta.json`/`sources-compat.json`).

Every leg is also individually dispatchable from the command line: `make promote TAG=... CHANNEL=... [FORCE=1]`, `make release-status`, `make release-verify`, `make release-publish-pypi/-packages/-docker/-cuda`, and `make release-attach` for stranded candidate runs.


tobocop2:
Closing to rethink the promotion model before committing to the channel-pointer approach.

PR #541, opened 2026-07-16
Refuse a second server on a busy data dir, and log signal-driven shutdowns

tobocop2:
## Problem

Nothing stops two `lilbee serve` processes from running against one data dir: the second overwrites server.port and spawns a second model fleet against the same models and vector store. A signal-terminated server also leaves no trace in server.log, so killed sessions read like crashes in diagnostics, and a supervisor that needs to replace a running server has no way to ask it to stop short of signalling its pid.

## Solution

`serve` holds an OS file lock next to server.port for its whole lifetime, plus one in `LILBEE_EXCLUSIVE_SCOPE` when a supervisor manages several data dirs. The kernel releases the locks on any death, and they are held until the fleet teardown finishes, so a successor can never start while the predecessor's models still occupy memory. A second serve waits up to fifteen seconds, then exits with a distinct code (3) and a message naming the holder.

`POST /api/shutdown` (token-authed) stops the server through the same path as an external SIGTERM. Shutdowns now log the received signal and the fleet teardown outcome, and the per-group stop grace drops to 2.5 s so the worst-case teardown stays inside the lock-handoff window.


tobocop2:
Folding this into a bigger feature: rather than refusing a second lilbee on a busy data dir, the second instance should join the running fleet as a client. The locks, shutdown endpoint, and teardown work here are the substrate for that, so this branch continues as the feature's base and the whole thing will ship as one PR.

PR #547, opened 2026-07-17
Share one inference engine across every lilbee process a user runs

tobocop2:
## Problem

Every lilbee process builds its own model engine. Open the TUI while a coding agent uses `lilbee mcp`, or run `lilbee serve` alongside either, and each process spawns a full llama-server fleet: identical weights load twice, and on small machines both fleets degrade or OOM.

A second `lilbee serve` on one data dir also silently overwrites `server.port` and the session token.

## Solution

One inference engine per OS user, shared by every lilbee process that user runs against the same model setup. Two accounts on one box get one engine each, since the slot lives in a per-user directory.

### How a process gets its engine

Each process runs a short ladder under a cross-process build lock.

It **binds** to the machine slot's running engine when that engine is healthy and serves the configured models. Binders spawn nothing and write nothing; they point at the recorded proxy ports.

It **builds** into the slot when the slot is empty. An engine that no live process is using is replaced in place, so leftovers never poison the slot.

It **overflows** to the config root's private dir only when the slot holds a live incompatible engine in active use. That is exactly the case of two genuinely different model setups running at once.

### What "compatible" means

The contract is the per-role models plus the bundled engine-build pin. lilbee version is deliberately not part of it: releases sharing an engine pin share an engine, while differing pins never run on a build they were not tested against.

### Lifetime

Membership is kernel-refcounted. Each process holds a user lock the OS releases on any death, and the last clean exit stops the engine, leaving the machine clean by default.

`keep_engine_warm` is the explicit opt-in for the engine to outlive lilbee. The idle TTL applies in every mode, so even a persistent engine releases weights when unused. `lilbee engine stop` remains the unconditional off switch.

### What this deletes

The old ownership machinery (detach/adopt, owner-liveness checks, sibling sparing, the ps-scan sweep) is removed outright.

### Server groundwork

`lilbee serve` gains a per-data-dir OS lock, plus an opt-in wider one that engages only when `LILBEE_EXCLUSIVE_SCOPE` names a directory, for supervisors that need at most one server across several roots. It also gains a token-authed `POST /api/shutdown` and shutdown logging, and the fleet teardown's SIGTERM grace drops from ten seconds to 2.5 on every path that stops a fleet, not only on serve.

---

## Hardening found by QA

Everything below was discovered while load-testing and benchmarking the shared engine, not planned up front. Each item is a capability or a correctness rule that was **missing** rather than broken by this branch.

### Expert offload was missing entirely

A sparse mixture-of-experts model activates a fraction of its weights per token, so its experts tolerate system memory far better than a dense model's layers would.

llama.cpp has supported offloading them since well before the version lilbee pins, and lilbee emitted none of it: no `--cpu-moe`, no `--n-cpu-moe`, no `--override-tensor`. Every model therefore had to fit VRAM outright, which is why large MoE coders appeared to need a datacenter card.

`cpu_moe` and `n_cpu_moe` now emit those flags, gated on the GGUF declaring routed experts so the flag is never a silent no-op on a dense model. No engine bump was needed; the pinned build already had this.

### The estimator did not know about offload either

gguf-parser was sizing models as though every expert were resident, so the planner budgeted against a footprint that never materializes and granted fewer batching slots than the real residency warranted.

gguf-parser accepts `--override-tensor` with the same pattern syntax llama.cpp uses. The estimate offloads what the launch offloads. The pattern is copied from llama.cpp rather than shared with it: the launch passes `--cpu-moe`/`--n-cpu-moe` and lets the pinned llama-server expand its own regex, while the estimator is handed our copy. The two move together only as long as that copy is kept in step with upstream, which the constant says where it is defined.

### The catalog called runnable models unrunnable

The fit chip that marks a model "won't run" was computed against GPU VRAM alone. With offload, a model larger than VRAM runs fine, so the catalog was discouraging pulls that now work.

The budget borrows system RAM when offload is configured, and only on hosts whose budget really is device memory. Apple unified memory, non-NVIDIA hosts, and CPU-only hosts already report system RAM, and adding it twice would invent capacity.

### Quantized KV could produce a command line that would not load

Quantized KV cache requires flash attention, and nothing checked the pairing. With `flash_attention` off and the default `q8_0` cache, the launch emitted `--flash-attn off` alongside `--cache-type-k q8_0`.

Both the launch and the estimator now fall back to f16 so their sizing agrees.

### The setup self-check disagreed with the fleet

The self-check built its own llama-server command line and re-derived the chat flags in a hand-written expression, so it carried that same unloadable pairing and knew nothing about offload. A user with offload configured would have watched the diagnostic fail on a configuration the fleet launches fine.

Both callers now share one set of flag decisions instead of two copies.

### Smaller fixes from the same passes

Catalog commands (`model pull/show/rm/browse`) no longer eager-start the fleet, so a download can never spawn a partial engine mid-pull. A chat request whose output reservation exceeds the served window is clamped instead of rejected, so agent clients that over-reserve keep working. The membership locks survive cross-thread release, multiple lilbee instances in one process, and thread-pool reuse.

---

## Acceptance test

A four-agent load harness ships with the branch (`tools/qa/shared-engine`): four opencode agents on one GPU box, each in its own project, all served by one engine. It exercises binding instead of building, VRAM against a single-agent baseline, task completion, and teardown leaving zero engines and zero locks.

## On the load and stability numbers

An audit of that harness found its invariant checks too weak to cite as evidence, so the throughput, VRAM-flat, and leak-free figures previously quoted here have been removed rather than restated. The VRAM check is one-sided, the documented "process counts return to baseline" invariant is not implemented, the one-engine assertions accept a range that admits the duplicate-engine regression they exist to catch, and the bind check samples before the agents send any requests. The published percentiles also were not reproducible from any committed script. Run output is machine-specific and is no longer version-controlled.

That work is tracked separately and lands before any number from this harness is quoted again.

An earlier soak run is still why two of the fixes above exist: on the pre-fix build, killing the llama-swap proxy orphaned its VRAM, built a duplicate engine in the overflow slot, and left the resident server erroring until restart. That reproduction stands on its own.

## Folded in from #572

From a deep-retrieval review of a file this branch owns (commit `68c3e504`; #572 closed in favour of this PR):

The server lifespan wrapped the embedding check in a try/except and logged "Embedding model validated" regardless of outcome. `validate_model()` is a bool predicate that cannot raise, and must not: search and chat degrade deliberately when no embedder is available. The except was dead and the success line was false whenever the model was missing. Its test stubbed a `RuntimeError` the real method cannot produce, so it asserted nothing.

Now branches on the returned bool, names the unavailable model in the warning, and reuses the services handle already built for the provider pre-load instead of calling `get_services()` twice. The test covers the false-return path.


tobocop2:
Added `68c3e504` directly to this branch (was PR #572, from a deep-retrieval review of files this branch owns):

The server lifespan wrapped the embedding check in a try/except and logged "Embedding model validated" regardless of outcome. `validate_model()` is a bool predicate that cannot raise, and must not: search and chat degrade deliberately when no embedder is available. So the except was dead and the success line was false whenever the model was missing. Its test stubbed a `RuntimeError` the real method cannot produce, so it asserted nothing.

Now branches on the returned bool, names the unavailable model in the warning, and reuses the services handle already built for the provider pre-load instead of calling `get_services()` twice. Test covers the false-return path.

PR #548, opened 2026-07-17
Add a blind, noise-calibrated retrieval eval harness

tobocop2:
## Problem
Retrieval quality changes have been graded by an ad-hoc kit living outside the repo, so every tuning decision starts from scratch and nothing gates regressions.

## Solution
`evals/retrieval` (repo-level, never shipped in the wheel) generates a question battery from any indexed library: topical questions from reservoir-sampled passages, known-item questions from sampled documents, and count questions with exact streaming-scan oracles. It answers the battery against two lilbee servers over the same index with per-row checkpointing so killed pod runs resume, judges answers blind under opaque ids with one arm duplicated to measure the judge's noise floor, and scores mechanically: per-dimension means with the noise floor as the error bar plus exact pass/fail for counts and document lookups. Streaming scans keep memory flat on very large libraries. Lint and format gates now cover the directory.


tobocop2:

tobocop2:

PR #549, opened 2026-07-17
Make document titles searchable in hybrid retrieval

tobocop2:
## Problem
The lexical index covers only chunk text, so BM25 can never match a document's title or filename; a query naming a document finds it only if the words happen to appear in the body.

Three adjacent defects in the same hybrid-search surface come along for the ride, all of which silently degrade retrieval. The FTS indexes were built without token positions, so a phrase query (raw user text reaches LanceDB as a phrase) errored instead of matching. Rank fusion weighed the vector and BM25 arms equally with no way to favor a strong dense embedder. And `ensure_fts_index()` runs `table.optimize()` on an existing index, which on a real-sized corpus can hit a LanceDB encoding bug and raise, leaving the index flag unset so every subsequent query silently fell back to vector-only with no error surfaced.

## Solution
Every chunk carries a nullable title (extraction metadata when available, cleaned filename stem otherwise) with its own FTS index, fused as a third reciprocal-rank arm at a configurable weight when `title_search` is on (off by default until the eval harness proves it). Existing chunk-FTS call sites now pin their column explicitly, because once a second FTS index exists an unpinned query silently searches every indexed column. Old indexes keep working untouched: schemas evolve in place and the title arm contributes nothing until a reindex populates titles. Source rows also persist extraction title, authors, and creation date.

Both FTS indexes are built with token positions so phrase queries match instead of erroring. `lexical_fusion_weight` (default 1.0, so equal-weight fusion is byte-for-byte unchanged) scales the BM25 arm down toward the dense arm. And an existing, queryable index is marked hybrid-ready before the best-effort `optimize()`, so an `optimize()` failure logs a warning instead of disabling hybrid search corpus-wide. A retrieval benchmark on BEIR confirmed the optimize bug: every hybrid config scored identically to the vector-only baseline until it was fixed.


tobocop2:

tobocop2:

PR #550, opened 2026-07-17
Widen retrieved passages with neighboring chunks

tobocop2:
## Problem
Retrieved chunks are prompted exactly as stored, so a hit landing mid-argument loses the sentences before and after it and the model answers from a fragment.

## Solution
With `neighbor_expansion` set (off by default), each selected chunk pulls up to N adjacent chunks per side from its own source and merges them into one contiguous passage, deduplicating the chunker overlap. Expansion runs after the token-budget fit and spends only leftover budget, so an original chunk is never dropped for a neighbor; citation numbering is untouched and page/line spans are recomputed truthfully. Indices claimed by a higher-ranked passage are never pulled twice, so no text repeats across passages.


tobocop2:

tobocop2:

PR #552, opened 2026-07-17
Add a standardized, preregistered retrieval benchmark harness

tobocop2:
## Problem
The retrieval work needs a comparison that survives scrutiny, not a homegrown judge a skeptic can wave away. That means standard metrics, public human-labeled data, and statistics on every claimed difference.

## Solution
`evals/benchmark` compares lilbee against another RAG system on public labeled datasets with the generator and embedding model held constant, so retrieval is the only variable. Tier 1 scores each system's ranked results against human relevance labels with pytrec_eval (nDCG@10, Recall@20, MRR@10) with no model in the loop, so the core numbers are exactly reproducible. Tier 2 layers RAGAS answer metrics plus the existing blind judge on top. Passage datasets carry native TREC qrels; the table and PDF datasets derive retrieval labels from human gold-evidence annotations through one documented function, disclosed in the frozen manifest. That manifest fingerprints the datasets, metrics, and models before a run and refuses a judge equal to the generator; paired-bootstrap confidence intervals and a randomization p-value gate every reported difference, and a difference whose interval crosses zero is reported as not significant.

The heavy scorers and dataset loaders stay in a standalone pod-install requirements file so installing them never moves the shipped lock, and they are imported lazily so the suite runs without them. The harness lives outside the package, so nothing here ships in the wheel.


tobocop2:

tobocop2:

PR #554, opened 2026-07-17
Fix a crash when the data root is set via LILBEE_DATA_ROOT

tobocop2:
## Problem

Setting the data root through the `LILBEE_DATA_ROOT` environment variable crashes at startup. The variable maps straight onto the `data_root` field as a string, and the config resolver then derives the documents, data, and lancedb directories from it with the `/` operator, which raises `TypeError: unsupported operand type(s) for /: 'str' and 'str'`. Every other way of setting the data root passes a `Path`, so this path was the only one that broke.

## Solution

Coerce `data_root` to a `Path` before deriving the child directories, and write the coerced value back so the field is consistently a `Path`. Coercing an existing `Path` is a no-op, so nothing else changes. Added a regression test that sets `LILBEE_DATA_ROOT` and asserts the child paths resolve without raising.

### DIFF
--- a/src/lilbee/core/config/model.py
+++ b/src/lilbee/core/config/model.py
@@ -966,7 +966,10 @@ class Config(BaseSettings):
             else:
                 local = find_local_root()
                 data["data_root"] = local if local is not None else default_data_dir()
-        root = data["data_root"]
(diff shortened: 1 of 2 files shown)


tobocop2:

tobocop2:

PR #556, opened 2026-07-18
Adaptive per-query fusion, on by default

tobocop2:
## Problem
The retrieval benchmark showed no single lexical fusion weight wins every corpus: on BEIR, SciFact wants a strong BM25 arm, NFCorpus a weak one, and FiQA none at all. A fixed `lexical_fusion_weight` is therefore always wrong somewhere, and the shipped equal-weight default significantly regressed FiQA (financial Q&A, where lexical term-matching mostly adds noise) by 0.056 nDCG@10.

## Solution
Scale the BM25 arm per query by how confident the vector arm is: a peaked dense ranking (a top hit standing clear of the field) downweights the lexical arm toward zero, a flat one keeps it. `lexical_fusion_weight` becomes the ceiling the rule scales down from, and `adaptive_fusion_margin` sets the vector-similarity margin at which the lexical arm is fully silenced.

On the same three datasets, adaptive fusion at margin 0.15 beats the fixed w=1.0 default on all three, keeps the significant NFCorpus (+0.0085, p=0.033) and SciFact (+0.0205, p=0.009) wins, and cuts FiQA’s regression to -0.018 (from -0.056 fixed, and better than the best hand-picked fixed weight). It is the strongest single retrieval policy the study found, so it ships on by default (`adaptive_fusion=true`, `margin=0.15`); set `adaptive_fusion=false` to pin the fixed weight. Full numbers and reproducible run files are in `evals/benchmark/RESULTS.md`.

Stacks on #549 (the `lexical_fusion_weight` knob it builds on).


tobocop2:

tobocop2:

PR #557, opened 2026-07-18
Retrieval quality: title-aware hybrid, adaptive fusion, context expansion, and structural filtering

tobocop2:
Rolls up the retrieval-quality work into one branch so it merges as a unit. Each section below also has its own PR for detailed history.

Defaults: adaptive fusion, the title arm, the structural filter, and context expansion are all off (opt-in), pending a benchmark run to validate them.

## Problem

- Retrieval matched chunk text only, so a query naming a document by title often missed it.
- FTS indexes with token positions overflow LanceDB's list encoding on `optimize()` for a large corpus, and an `optimize()` failure silently dropped every query to vector-only.
- A chunk hit mid-argument lost the sentences around it.
- No single fixed lexical/vector fusion weight won on every corpus.
- Document-structure chunks (TOCs, cover pages) diluted context precision on a government-document corpus.
- Every prefilter (`source`, `chunk_type`, `chunk_concepts.chunk_source`) was a full-table scan.
- `LILBEE_DATA_ROOT` delivered a string and crashed path resolution (`str / "documents"`).

## Solution

- **Title arm (#549).** Chunks carry a nullable title (extraction metadata or cleaned filename) with its own FTS index, fused as a third reciprocal-rank arm. Off by default (`title_search`); the column and index are always built at ingest. `_sources` gains `title`/`authors`/`created_at`, migrated in place on first open.
- **FTS fixes (#549).** Indexes are built positionless and queried through `MatchQuery`, so a query is never parsed as a phrase. Exact-phrase matching is not supported — deliberate, for index stability. Each arm pins its own column so chunk and title don't double-count. An existing queryable index is marked ready before the best-effort `optimize()`, so failure logs a warning instead of disabling hybrid search. `lexical_fusion_weight` makes the BM25 arm tunable.
- **Context expansion (#550).** Selected chunks pull adjacent same-source chunks into one passage, deduplicating the chunker's overlap, spending only the budget left after the originals fit. Off by default (`neighbor_expansion=0`).
- **Adaptive fusion (#556).** Scales the BM25 arm per query by vector-arm confidence. On by default.
  **Evidence caveat:** the BEIR numbers that justified this default came from the eval harness later audited in PR #569, which found 63 measurement defects including a wrong MRR metric and a denominator that dropped unanswered queries. After correction only 13 of 36 comparisons survive, the adaptive-fusion NFCorpus win is withdrawn, and every fusion weight loses to dense on FiQA at adjusted p <= 0.0016. Fresh runs under the corrected harness need indexed BEIR corpora and have not been done. The structural finding that no fixed weight wins every corpus still holds, which is the argument for adapting per query, but treat `adaptive_fusion=true` as unvalidated pending a re-run rather than as a measured win. The fused score is now a within-query signal, not comparable across queries — treat `min_relevance_score` as a coarse floor.
- **Structural filter.** Drops unambiguous TOCs and classification-banner cover pages from the `top_k*2` buffer, never dropping a lexically hit or top-ranked page. Off by default (`filter_structural_chunks`): the same A/B found it net-negative overall, since the cover-page heuristic also caught short banner-carrying body pages the answer needed.
- **Scalar indexes.** BTree on `source`, Bitmap on `chunk_type`, BTree on `chunk_concepts.chunk_source`. Built on next sync or first search, serve-only stores included. Additive on-disk change; drops no data, needs no re-ingest.
- **Config crash (#554).** The env var value is coerced to `Path` up front.

The eval and benchmark harness is not here — it lives on `evals/retrieval-harness` so this PR stays to lilbee's core.


tobocop2:
Added `2dffa93e` directly to this branch (was PR #573, from a deep-retrieval review of files this branch owns):

- Ingest pipeline and entity backfill now use the public `load_spacy_pipeline` instead of importing `_ensure_spacy_model` across packages. `concepts/graph.py` keeps the private call since it is in the same package.
- HyDE gets its own `HYDE_MAX_TOKENS`; it was sharing `EXPANSION_MAX_TOKENS` with query-variant expansion, so retuning one silently retuned the other.
- Removed `COMPACT_PROMPT`'s `{previous}` slot, its window reservation, and the `previous_summary` parameter threaded through `plan_compaction`, `batch_overflow`, and `_summarize_batch`. Every caller passed an empty string, so it never rendered and only made batches smaller than needed.

PR #558, opened 2026-07-18
Retrieval eval + benchmark harness (tooling)

tobocop2:
The blind retrieval eval harness (`evals/retrieval`) and the preregistered lilbee-vs-baseline benchmark (`evals/benchmark`), split out of the core retrieval work (#557) so the harness tooling lives on its own branch instead of cluttering lilbee's core.

Runs against a deployed lilbee and imports the public API only, so nothing in `src/` depends on it. Includes the Qwen3-0.6B benchmark results (`evals/benchmark/RESULTS.md` + gzipped TREC run files) that validated the retrieval features in #557.


tobocop2:
Not merging the harness into main. It stays on its own standalone branch (evals/retrieval-harness) as separate tooling; closing this PR so it never targets main. Branch and history remain intact.

PR #569, opened 2026-07-20
Score with ir_measures and grade with RAGAS instead of hand-rolled metrics and judging

tobocop2:
## Problem

The harness scored, graded, and loaded everything itself. That was the bug, not a style preference.

The metric layer computed MRR@10 by truncating each run to depth 10 and rescoring the survivors, which is a different computation from scoring the full run under a cutoff. It disagreed with trec_eval on 99 of FiQA's 648 topics, in both directions, and the published FiQA number was inflated. nDCG@10 and Recall@20 were unaffected on all 15 arms.

The tier-2 judge was a second bespoke eval framework: a prompt template, a last-brace-wins regex over the judge's reply, a hand-written range check, and a three-attempt retry loop against a greedy decoder that returns the same answer every time.

The baseline was confounded. `dense` was the only arm with `title_search` off and the baseline for every `-vs-dense` comparison, so no delta was attributable to the fusion weight the ablation existed to measure.

The manifest recorded no build provenance, so its runs could be rescored but not reproduced. Datasets came from an unversioned zip URL, so a republished corpus moved every number while the frozen fingerprint stayed identical.

And the pinned `ragas==0.2.14` did not import at all on a fresh install. `ragas.llms.base` imports `langchain_community.chat_models.vertexai` unconditionally, langchain-community 0.4 removed that module, and ragas declares no upper bound. The requirements file pinned the scorer and let its dependency float.

## Solution

Metrics are `ir_measures`, so the cut depth is part of the measure (`RR@10`) and nothing truncates a run. Statistics are `scipy`: percentile bootstrap, paired sign-flip permutation test, Benjamini-Hochberg. The new layer reproduces the reference exactly on the committed run files: nDCG@10 0.4033, Recall@20 0.5802, MRR@10 0.4700.

`ranx` was evaluated and rejected. It cuts depth correctly, but it ships its own nDCG variant and scores 0.4060 where trec_eval says 0.4033. Published BEIR baselines are trec_eval numbers, and a harness whose numbers cannot be compared to them is worth less than one that is merely correct.

Grading is RAGAS' `DomainSpecificRubrics`, which owns the prompt, asks for a typed result through instructor instead of parsing prose, and retries what does not validate. What stays is the part no library ships: the three dimensions, their rubric text, and the two equivalent presentations of each rubric that make the duplicate-arm noise floor measurable. Both backends decode greedily at temperature 0, so an identical rubric would produce an identical prompt and a noise floor of zero by construction.

Datasets load through `ir_datasets`, which drops `beir` and `datasets` and replaces the unversioned URL with an id that pins and checksums a published copy. A new `fetch` command materializes corpus, queries, and TREC qrels, so what ships beside a report is what every other IR tool reads.

The baseline confound is gone: `title_search` gets its own arm. On fiqa, whose BEIR corpus carries no title field, that arm is a deliberate A/A null control, and a significant difference there is a harness defect rather than a finding. `SystemProvenance` records the commit and index parameters, and `require_reproducible` refuses to publish without them.

Along the way the retrieval report was labelling every mean `(0-2)` as a literal while the judge now grades 1-5. The scale is derived from the constants the judge uses, and a guard fails the run if a RAGAS release moves it. RAGAS telemetry is opted out at import: a benchmark should not make outbound calls its manifest does not declare.

## What this does not yet establish

Tier 1 has been verified numerically against the reference. Tier 2 has not been run end to end against a live judge, so the RAGAS binding is covered by construction tests against the real API and by stubs, not by a scored run. The old reports and artifacts were deleted rather than corrected, since they came from an unrecorded build under the confounded design; every number has to be regenerated before anything is published.


tobocop2:
Superseded by #581, which targets main directly.

This branch was stacked on `evals/retrieval-harness`, which had drifted behind main and still carried the previous harness' published artifacts. Those came from an unrecorded build under a confounded baseline, and from a metric layer whose MRR@10 disagreed with trec_eval on 99 of FiQA's 648 topics, so they were deleted rather than corrected. #581 is the same work on a clean branch off main, plus the run-provenance capture.

PR #572, opened 2026-07-20
Fix the server startup embedding check reporting success when the model is missing

tobocop2:
## Problem

The server lifespan wrapped the embedding check in a try/except and logged "Embedding model validated" regardless of the result. `validate_model()` is a bool predicate that cannot raise, and must not: search and chat are built to degrade when no embedder is available. The except was dead code and the success line was false whenever the model was missing.

Its test stubbed a `RuntimeError` that the real method cannot produce, so it passed without exercising anything.

## Solution

Branch on the returned bool, naming the unavailable model in the warning, and reuse the services handle already built for the provider pre-load instead of calling `get_services()` a second time. The test now covers the false-return path and asserts the warning.

### DIFF
--- a/src/lilbee/server/app.py
+++ b/src/lilbee/server/app.py
@@ -113,15 +113,18 @@ async def _lifespan(app: Litestar) -> AsyncIterator[None]:
     inject_provider_keys()
 
     try:
-        get_services()  # pre-load all services (provider, embedder, etc.)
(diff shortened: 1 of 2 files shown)


tobocop2:
Folded into fix/serve-data-dir-singleton directly as 68c3e504 to keep the PR count down. Closing this in favour of #547.

PR #575, opened 2026-07-20
Lock the wiki writers that were missing it, key ingest locks on the real source

tobocop2:
Stacked on #574, which is stacked on #570. Review those first; the diff here is only the three follow-up commits.

## Problem

The wiki build mutex exists because concurrent writers corrupt the pages, the index and the log. Build, update and synthesize took it. Prune, which archives and flags pages, and draft-accept, which overwrites a published page and re-indexes its chunks, did not. Both refresh `index.md`, the file the lock's own comment names.

The ingest lock registry reduced every key to its basename. That is right for `/api/add`, where `copy_files` flattens into `documents_dir`, but uploads keep their relative layout, so `src/util.py` and `tests/util.py` are different files that were serialized against each other and reported as contending.

The SSE cleanup cancels its in-flight future and awaited it under `contextlib.suppress(BaseException)`, which also covers `KeyboardInterrupt` and `SystemExit`, so a Ctrl-C landing on that await was discarded.

## Solution

Prune and draft-accept take the build mutex.

The ingest caller derives the lock identity, because how a path maps to a stored source differs per surface. `add` passes the basename explicitly; uploads pass the relative path they actually write.

The SSE cleanup catches `CancelledError` and `Exception` by name, and logs an upstream failure instead of dropping it silently.

## Notes for review

The serialization test drives the route handlers directly. Going through `AsyncTestClient` does not work: it serializes the requests it is given, so such a test passes whether or not the lock is there. The existing build-vs-build test has that flaw and is filed separately.

The bead behind the SSE change also claimed the suppression let a cancelled request task resume as if it had never been cancelled. That does not hold: `contextlib.suppress` inside a `finally` only suppresses exceptions raised within its own block, so a `CancelledError` already unwinding keeps propagating. Verified against both versions. Only the `BaseException` width was real, so the fix is just the narrowing.

### DIFF
--- a/src/lilbee/runtime/ingest_lock.py
+++ b/src/lilbee/runtime/ingest_lock.py
@@ -49,16 +49,25 @@ class IngestLockRegistry:
 
     @staticmethod
     def canonical_source_name(p_str: str) -> str:
-        """Match the basename ``copy_files`` writes under ``cfg.documents_dir``."""
+        """Basename of *p_str*, for callers that flatten into ``documents_dir``.
+
+        ``copy_files`` writes server-side paths by basename, so /api/add locks
+        on that. Uploads keep their relative layout and pass the relative path
+        instead, which is why the caller picks the key rather than this class.
+        """
         return Path(p_str).name
 
-    async def acquire(self, paths: list[str]) -> tuple[list[tuple[str, asyncio.Lock]], list[str]]:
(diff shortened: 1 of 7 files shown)


tobocop2:
Consolidated into #570, which now carries all 23 commits against main.

PR #588, opened 2026-07-21
Make document removal folder-aware and durable across all surfaces

tobocop2:
## Problem

Removing a document meant three different things depending on where you asked. The TUI's `/delete` wrote a skip-marker so the next sync wouldn't re-ingest a file left on disk, but the CLI, MCP, and REST removes called the store directly and skipped that step. So any removal that kept the file on disk (the default, without `--delete`) silently undid itself on the very next sync.

There was also no way to drop a whole directory in one step. `remove` matched exact source names only, so clearing an added repo meant naming every file under it. A plugin user who synced a GitHub repo and later wanted the stale content gone had to delete the folder by hand in Obsidian.

Underlying both: `add` copies files into the vault, so the copy is a point-in-time snapshot. Nothing in the docs said so, which is why the tracking expectation formed in the first place.

## Solution

Route every user-facing removal (CLI, MCP, REST, the library API, and the TUI) through `remove_documents_durably`, so a removal sticks regardless of entry point. The wrapper now also expands a folder name to every source indexed beneath it, matched on whole path segments (`myrepo` covers `myrepo/a.py` but never `myrepo-2/x`; a trailing slash forces the folder reading when a same-named file also exists). The CLI prompts before a folder removal and takes `--yes` to skip; the prompt is auto-skipped in json mode and for the non-interactive callers.

The two internal sync-reconciliation calls stay on the direct store path on purpose: the file is already gone from disk there, so writing a skip-marker would be wrong.

Docs now state that `add` copies into the vault, that the copy is a snapshot, and that dropping a stale directory is a `remove`, not an edit at the source.

No new TUI screen: the Status view already lists every source, so folder removal rides the existing `/delete` and `lilbee remove` surfaces rather than a dedicated panel.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -67,6 +67,12 @@ After that you're in chat. `/add` indexes documents, `/crawl` indexes a website
 Press `/add` to add files, directories, or web pages. Paths tab-complete; the
 job runs in the Task Center, so you can keep chatting while indexing happens.
 
+`add` copies the files into lilbee's own documents directory, so the copy is a
+snapshot taken at that moment. Editing the originals later has no effect until
+you add them again, and files you delete at the source stay indexed until you
+remove them. To drop a whole directory you added, remove it by name with
+`/delete myfolder` (or `lilbee remove myfolder`).
+
 If a file with the same name is already indexed, `add` skips it. To re-index in
 place, remove the document first with `/delete name`, or pass `--force` from
 the CLI.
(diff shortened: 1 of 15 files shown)


tobocop2:
Superseded by the symlink-in-place ingest work. The folder-aware removal here (whole-segment folder expansion, the multi-document confirmation prompt) is carried into that branch and extended with glob-pattern removal. The destructive parts of this PR (--delete file deletion and empty-directory pruning) are dropped there on purpose: with sources linked into the knowledge base rather than copied, remove is now index-only and never touches source bytes on disk.

PR #594, opened 2026-07-22
Capture the Windows wedge's native frame with an external py-spy watchdog

tobocop2:
## Problem

- The faulthandler watchdog named the wedged test (a `SessionStore` file-I/O volume test) but produced no stack: a continuous GIL hold in native code blocks even faulthandler's own dump.
- So we know *which* test wedges, not *where* it blocks.

## Solution

- External py-spy watchdog (`scripts/qa/hang_watchdog.py`): wraps the test command, and on a total output stall dumps every worker's native stack (`py-spy dump --native --nonblocking`) then kills the run, so a wedge fails fast with the frame instead of hanging to timeout-minutes.
- py-spy reads the target from outside, so it dumps a GIL-frozen worker where faulthandler can't; no admin needed for a same-user process on Windows.
- Wired on the Windows jobs only; passthrough (no-op) elsewhere. Dumps land in the dir the existing surface step prints.

Verified locally: stall-detect/dump/kill fires; transparent (exit code + output pass through) when armed without a stall. The dump content is only confirmable on Windows CI (py-spy needs root on macOS).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,15 @@ jobs:
       - name: Install dependencies
         run: uv sync --locked
 
+      - name: Install py-spy (Windows hang watchdog)
+        # External native profiler for the Windows wedge: it reads the frozen
+        # worker's memory from outside, so it captures a native stack where the
+        # in-process faulthandler watchdog cannot (a continuous GIL hold blocks
+        # even faulthandler's own dump). No admin needed for a same-user process
+        # on Windows. Into the venv so `uv run` finds it on PATH.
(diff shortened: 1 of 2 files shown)


tobocop2:
Closing: the wedge this was built to diagnose (the SessionStore fsync storm) is now root-caused and fixed test-only in #598, without needing py-spy's native frame — which never worked on Windows py3.13 anyway (py-spy could not read the interpreter version). The faulthandler watchdog already merged in #592 names any future wedged test. If a genuinely new 2h hang ever appears, an external stall-killer is a small, well-understood change to re-add then, rather than carried speculatively now.

PR #611, opened 2026-07-24
Coalesce concurrent ingest embeds into full batches

tobocop2:
## Problem

Bulk ingest of single-chunk documents (MS MARCO passages, tweets, log lines) issued one `/v1/embeddings` request per document. Each file runs on its own ingest thread, so hundreds of one-passage requests were in flight at once, but their per-request Python work (request build, JSON parse, base64 decode, vector materialization) is GIL-serialized at a few milliseconds each. That capped aggregate throughput near ~150 docs/sec no matter how fast or numerous the GPUs were: the cards starved because the client could not emit requests fast enough, not because they could not compute. `embed_batch_sequences` never helped because each caller only ever handed the client one sequence.

## Solution

Add a coalescer (`providers/fleet/embed_coalescer.py`) that merges concurrent per-file embed calls back into full batches (up to `embed_batch_sequences`), so the fixed per-request cost is paid once per batch instead of once per passage. A single batcher thread groups queued requests; a bounded pool (a couple of batches per replica) dispatches them concurrently through the existing failover routing, so the batches still fan out across replicas.

It is scoped to bulk ingest via a `coalesce_embeds()` context in `sync()`, so the interactive query and chat paths keep their direct, latency-free dispatch. One bad input is retried alone so it cannot fail its batch-mates, preserving the per-file failure isolation ingest had before requests were merged.

---
Stacked on #609 (streams the ingest plan so embedding starts before the corpus is hashed); merge that first. The two are complementary: #609 removes the plan/hash idle preamble, this removes the steady-state dispatch ceiling.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -100,9 +100,16 @@ CLI also accepts `--model` / `-m` for chat model, `--data-dir` / `-d`, `--ocr-ti
 
 ### Test-Driven Development
 - **100% test coverage required** — enforced by `pytest-cov` with `fail_under = 100`
+- **Check coverage of what you changed before pushing.** Lint, typecheck and a targeted test file are all green in seconds and none of them can see an untested new branch; CI then fails with `Coverage failure: total of 99 is less than fail-under=100` and *every test passing*, which reads like an infra problem and is not. Scope the same check locally instead of skipping it:
+  ```bash
+  uv run coverage run -m pytest tests/test_<suites touching your change>.py
+  uv run coverage report -m --include='*<changed_module>*'
+  ```
 - Write tests BEFORE or alongside implementation, not after
 - Every public function MUST have at least one test
 - **Every test must assert something meaningful** — no zero-assertion coverage padding
+- **A test that cannot fail is not a test.** Before accepting an assertion, mutate the code it guards in your head: if the test still passes, it is decoration. The common shape is a bound asserted against inputs that could never violate it (`assert all(len(b) <= 8 ...)` where every input is a single item), and an error/cleanup path tested without building its precondition.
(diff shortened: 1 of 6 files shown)


tobocop2:
Throughput validation on 2xA40 (50k MS MARCO subset) shows a **regression, not a gain**: baseline (main) 47.1 docs/sec vs coalescing 38.8 docs/sec (0.82x). Both landed the same 49,907 rows, so the measurement is valid.

Why: 2xA40 is GPU-bound for this workload (baseline ran at ~97% GPU util, far below the ~150 docs/sec per-request dispatch ceiling, which only appears at high card counts like 8xA40 where aggregate GPU capacity exceeds it). When the GPU is already the bottleneck, coalescing only adds cost -- coarser batching and lower server concurrency than the baseline's high-concurrency single-passage stream.

Two things must be true before this is mergeable, and neither is yet:
1. Confirm the gain on 8xA40 -- the regime where dispatch is actually the bottleneck. The 2-card test cannot validate this fix.
2. Make coalescing conditional so it never regresses the GPU-bound / few-GPU / multi-chunk-document case. As written it is always-on during ingest, which would slow those deployments down.

Converting to draft until both are addressed.

tobocop2:
Re-tested in the correct dispatch-bound regime (0.6B embedder on 4xA40, baseline pinned at the ~150 ceiling): baseline 147.5 docs/sec vs coalescing 82.1 docs/sec = **0.56x**, with the conditional gate engaged. Coalescing regresses ~45% even here, and GPU util never exceeded 2 of 4 cards -- it collapses the baseline's cross-card parallelism (many concurrent single-passage requests) into a few large batches that pile onto 1-2 replicas. The hypothesis that fewer/larger requests raise throughput is falsified. Keeping this draft; not a mergeable direction as implemented.

PR #623, opened 2026-07-26
Read the engine mirror from the wheel builds and stop storing engines twice

tobocop2:
## Problem

The first `warm-engine-cache` run on main after #621 went green but rebuilt every CUDA engine from source:

```
88min  (windows-2022, cu124)
69min  (windows-2022, cu125)
67min  (ubuntu-22.04, cu124)
62min  (ubuntu-22.04, cu125)
```

The mirror already held all four. `build-multigpu` publishes to the mirror but never reads it: it has its own inline `actions/cache` steps instead of going through `bundle-llama-server`, which is where the read path lives.

Separately, a built engine was written to both the mirror and the Actions cache. The cap was back to **9.90 GiB of 10 GiB** with 12 main-scoped engine entries totalling 2.56 GiB, before `warm-executables-cache` had created a single Nuitka object cache. Those would have evicted on arrival.

## Solution

`build-multigpu` resolves mirror, then cache, then source build, same order as the executable builds. All nine downstream toolchain-install and build guards move from the raw cache-hit to the combined decision, so a mirror hit skips the CUDA toolkit install too.

Engines are mirror-only on write. Object caches want per-file granularity and have to be a cache; an engine is one immutable blob and does not, so it should not compete for the cap. Reads still fall through to the Actions cache, so engines cached before the mirror existed are still used.


tobocop2:
Merged as part of #624, which was stacked on this branch and contains commit acfd07df verbatim. Splitting them at merge time would have conflicted: the review hardening touched files #624 renamed (engine_mirror.sh -> ci_mirror.sh), so cherry-picking it back here did not apply cleanly. Closing rather than leaving a PR whose diff against main is now a pure reversion.

PR #626, opened 2026-07-26
Let ingest workers write their own chunks

tobocop2:
> Stacked on #618. Review that one first; this branch contains it.

## Problem

Multiprocess ingest drains every worker's chunks (~16KB of vector each) over IPC to one parent that builds Arrow and writes single-threaded. That drain is the ceiling: 8xH200 sat at ~220 docs/sec with the cards at ~50% while one card alone runs 57.7 at 93%.

## Solution

Workers commit their own chunks as lance fragments straight to the shared dataset (concurrent appends don't conflict); the parent keeps sources, side tables and reconciliation. A row watermark captured before the workers spawn keeps cleanup exact: a re-ingested source's old rows are deleted at flush, the run's own appended rows survive, and crash orphans self-heal next sync.

Active only for multiprocess runs with the new `bulk-ingest` extra (pylance 9); single-process and interactive syncs keep the atomic single-writer path. Verified against real pylance + lancedb 0.33: racing appends and deletes lose no rows, counts reconcile, commit latency flat. Stays on lancedb 0.33; the pre-Haswell channel has no 0.34 compat wheel yet.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -218,6 +218,7 @@ The codebase has a small set of `try: import X except ImportError:` patterns for
 | `litellm` | `lilbee.providers.litellm_sdk.litellm_available()` | `lilbee[litellm]` | SDK provider, settings TUI |
 | `crawl4ai` | `lilbee.crawler.crawler_available()` | `lilbee[crawler]` | Web crawler |
 | `graspologic_native` | `lilbee.retrieval.concepts.nlp.concepts_available()` | `lilbee[graph]` | Concept-graph clustering |
+| `pylance` (`lance`) | `lilbee.data.ingest.fragment_writer.fragments_available()` | `lilbee[bulk-ingest]` | Worker fragment writes in bulk multiprocess ingest |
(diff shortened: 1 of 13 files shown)


tobocop2:
Closing: this regresses ingest throughput.

A/B on 4×A40, 0.6B, procs=4 (both runs completed 60k/60k):
- parent (single-writer): 214 docs/s @ 64% GPU
- fragment (per-worker): 167 docs/s @ 49% GPU

Same regression on 8×H200/8B (~131 vs ~220). It reproduces across both models and both GPU tiers, so it's structural, not a tuning knob.

Ruled out locally: commit overhead (5.5k–140k rec/s) and the Arrow build (13k–17k rec/s). Both are far above the ~220/s needed, so neither is the bottleneck.

Root issue: the parent writer was never the constraint. The ~220 plateau is the shared embed fleet's admission ceiling. Moving writes into workers breaks the continuous embed pipelining and feeds the GPUs worse — wrong bottleneck.

The multiprocess base (#618) stays; that one is a real win.

PR #638, opened 2026-07-29
Partitioned multi-host ingest: shard the corpus across pods, merge to one index

tobocop2:
## Problem

The merge concatenated every table across the shard databases with no checks.

- A missing shard merged into an index quietly short a slice of the corpus.
- A shard embedded by a different model or dimension merged anyway.
- `_meta` ended up with one row per shard, so store identity depended on which row a reader happened to pick.
- Every table was read with `to_arrow()` and concatenated, holding the whole corpus in memory. That is ~140GB at 8.8M x 4096-dim.

Two more defects made a sharded run diverge from a single-host one.

- `ingest.sh` bucketed documents by the shard-local counter. The bucket is part of the source path lilbee stores, so the same passage landed at a different path depending on how many shards ran.
- `checkpoint.sh` divided its milestones by the global corpus size, so a 2-shard host topped out near 50% and never pushed a final checkpoint. Every shard also uploaded to one rolling slot, so shards clobbered each other and a restore could return another shard's index.

### Verify before writing

Each shard writes a `shard_manifest.json` when its ingest finishes: which slice it is, the embedder identity read back from its own `_meta` row, and its final row counts.

The merge refuses unless all four hold.

| check | catches |
|---|---|
| completeness | a missing or duplicated shard |
| identity | mixed embedder, dimension or schema version |
| intactness | a shard truncated after its ingest |
| schema | shards built by different lilbee versions |

Refusal happens before the output directory is created, so nothing half-written is left behind.

### Stream instead of materialising

Rows move shard-to-merged through LanceDB's own `RecordBatchReader`.

Held Arrow bytes drop from the whole corpus to a single batch: 394MB to 4.1MB, measured on a 393MB corpus.

### Three kinds of table

- **concat** — `chunks`, `_sources`, `_page_texts` and friends. Shards hold disjoint doc-ids, so this is a plain union with no dedup.
- **singleton** — `_meta`, `_entity_schema`. One row by contract, so they are verified and written once.
- **corpus-wide** — `concept_*`. Cluster ids only mean something inside the shard that assigned them, so these are refused. Partitioned runs disable the concept wiki.

Table names come from lilbee's constants rather than string literals, so a rename cannot silently move a table between classes.

### Indexes are rebuilt, not unioned

BM25 IDF depends on corpus-wide document frequencies. A shard's FTS index encodes statistics for its slice only, so unioning them would mis-rank with no error anywhere.

The merge drives lilbee's own `ensure_vector_index` and `ensure_fts_index` over the merged table instead.

### Shard layout and checkpoints

`materialise.py` moves out of the `ingest.sh` heredoc and buckets by the global passage index, so a sharded run lays documents out exactly as a single host would.

`checkpoint.sh` and `restore.sh` now share one `shard_env.sh` for the slot path and the shard's share of the corpus. Those two have to agree or a restore returns the wrong index.

## How the merge runs

```mermaid
flowchart TD
    A["read manifests, row counts, _meta"] --> B{verify}
    B -->|"missing shard, wrong embedder,<br/>lost rows, schema drift"| R["refuse<br/>nothing written"]
    B -->|ok| C[classify tables]
    C --> D["concat<br/>chunks, _sources, _page_texts"]
    C --> E["singleton<br/>_meta, _entity_schema"]
    C --> F["corpus-wide<br/>concept_*"]
    F --> R
    D --> G["stream batches<br/>shard to merged"]
    E --> H[keep one row]
    G --> I["rebuild ANN + FTS<br/>with lilbee's builders"]
    H --> I
    I --> J[merged index]
```

## Coverage

`evals/` carries no coverage gate, so coverage was run over the changed modules directly. It sat at 77%, with the gaps on the refusal branches — the same defect class as having no guardrail at all.

`merge_shards` is now at 85%. The remainder is CLI wiring.

## Not yet validated

A single-host versus sharded-and-merged comparison on real hardware is the gate before a full run.


tobocop2:
Closing: this is benchmark/ingest tooling, not a change to lilbee itself, so it does not belong on main. It lives on feat/partitioned-ingest, which is the only branch carrying evals/infra and already has the work.

PR #641, opened 2026-07-30
Give each lilbee process its own block of engine ports

tobocop2:
## Problem

Two lilbee processes starting at the same moment could get overlapping engine ports, leaving one worker's client talking to another worker's engine.

- The picker binds a probe socket and closes it; llama-swap binds member ports later, on first request.
- The search start was spread by pid, so consecutive pids start **one port apart**.
- A fleet takes its ports **contiguously** (proxy + one per member).
- Starts one apart therefore overlap on all but one port.

Measured on 2xH100, one lilbee pinned per GPU, 20k passages, changing only the launch stagger:

| launch | docs/s | per-card util |
|---|---|---|
| simultaneous | 53.9 | 8% / 87% |
| 15s stagger | 98.5 | 90% / 88% |

## Solution

Each process takes a whole **block** of the window, not an offset into it.

- **Block = 64 ports.** A fleet needs 1 proxy + 1 per member, and embed/vision replicate per GPU, so 64 covers a 30-GPU host. Narrower, and a wide fleet spills into the next block.
- **Window 200 -> 8192 ports** to hold 128 blocks, still below every default ephemeral floor (Linux 32768, macOS 49152).
- **Fallbacks unchanged:** a full block walks on into its neighbours; an unknown ephemeral range (Windows) still lets the OS choose.

| launch | docs/s | per-card util |
|---|---|---|
| simultaneous, with fix | 100.0 | 95% / 94% |

Faster than the stagger because the stagger itself cost 15s of dead startup per worker.

Caveat: two pids congruent mod 128 still share a block. Consecutive pids, which is what any launcher produces, never collide.

### DIFF
--- a/src/lilbee/providers/fleet/swap_manager.py
+++ b/src/lilbee/providers/fleet/swap_manager.py
@@ -549,7 +549,12 @@ def _ephemeral_range() -> tuple[int, int] | None:
 # Where lilbee looks for engine ports when the kernel's ephemeral range is known.
 # Above the registered-service crowd, below every default ephemeral range.
 _PORT_SEARCH_FLOOR = 20000
-_PORT_SEARCH_ATTEMPTS = 200
+_PORT_WINDOW_SPAN = 8192
+# Block width. Each process searches one block, so concurrent lilbees hold
+# disjoint ranges. A fleet takes one proxy port plus one per member, and embed
+# and vision replicate per GPU, so 64 covers a 30-GPU host; a wider fleet spills
+# into the next block.
+_PORT_BLOCK = 64
 
 # Ports handed to a child that has not bound them yet. llama-swap binds a member
 # port only on that member's first request, so the probe socket is long closed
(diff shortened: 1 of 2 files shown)


tobocop2:
Validated at width 8 on an 8xH100 pod, with the committed 64-wide artifact rather than the wheel the description's 2-card numbers came from.

**Ports each engine actually bound**, eight workers launched simultaneously with no stagger:

| bound ports | offset from 20000 | / 64 | block |
|---|---|---|---|
| 20000 / 20001 | 0 | 0.0 | 0 |
| 20064 / 20065 | 64 | 1.0 | 1 |
| 27360 / 27361 | 7360 | 115.0 | 115 |
| 27488 / 27489 | 7488 | 117.0 | 117 |
| 27744 / 27745 | 7744 | 121.0 | 121 |
| 27808 / 27809 | 7808 | 122.0 | 122 |
| 27936 / 27937 | 7936 | 124.0 | 124 |
| 28000 / 28001 | 8000 | 125.0 | 125 |

Every port lands exactly on a 64-boundary and all eight blocks are distinct; each worker took its block's first two ports (proxy + embed member). `_PORT_BLOCK=64 window=8192` confirmed in the installed wheel.

All eight cards held their own engine at ~9868 MiB. A worker that adopted a sibling's fleet shows up as a card at 0 MiB, which was the original failure signature and is absent here.

**Throughput, 8 workers on 8xH100, 80k passages, no stagger:**

| N workers | docs/s | per-card util |
|---|---|---|
| 2 | 111.9 | 94 / 94 |
| 4 | 205.1 | 93 / 92 / 92 / 92 |
| 8 | 285.7 | 87-91 across all eight |

Steady-state per-card rate is ~61.5 docs/s at every width; the aggregate at N=8 is held down by per-run fixed cost (~117s of engine loads and index builds against a 280s run), not by contention. Baseline for the same corpus was 161.9 docs/s for the best single-process configuration.

PR #666, opened 2026-08-02
Make wiki lint, prune and dataset export stoppable

tobocop2:
## Problem

`lint_all` and `prune_wiki` walk every wiki page and cost a store lookup each, so a large wiki takes long enough to want stopping. Neither took a cancel token on any surface.

## Solution

Both stop at the next page and report what they found, matching how the build stops at a source boundary. Prune still reconciles and finalizes what it collected, so a cancelled pass leaves the wiki consistent rather than half-scanned. Wired at MCP through the published token and at the CLI through SIGINT.

Dataset export is stoppable too. The jsonl writer already ran a batch at a time, and the parquet writer's docstring recorded that driving a `ParquetWriter` row group by row group is byte-identical to `write_table`; taking that path gives both formats a boundary. A cancelled write removes the partial file. The table build ahead of it is a single columnar scan with no Python boundary, and deliberately so, so the stop lands on the write.

## Verification

| Check | Result |
|---|---|
| `make lint` / `format-check` / `typecheck` | 0 / 0 / 0 |
| Scoped coverage (`wiki/lint.py`, `wiki/prune.py`, `cli/commands/wiki.py`, `mcp_server.py`) | 100% |
| Mutant: lint ignores the token | test fails |
| Mutant: prune ignores the token | test fails |
| Mutant: CLI drops the token (build, synthesize) | test fails, each case |
| Mutant: SIGINT handler never sets the token | test fails |
| Mutant: SIGINT handler stays installed | test fails |
| Mutant: parquet writer ignores the token | test fails |
| Mutant: jsonl writer ignores the token | test fails |
| Mutant: partial export file not removed | test fails |
| Parquet bytes vs `write_table` (table with rows) | byte-identical, pinned by test |
| Parquet with an empty table | footer differs (no row groups); still reads back as 0 rows, and the build refuses an empty export before the writer |
| Real MCP client cancels `wiki_lint` over the HTTP mount | **stopped at 14,507 of 80,000 issues** |
| opencode (free zen model) calls lilbee tools over the mount | `settings_list` and `wiki_lint` both dispatched |

The end-to-end run used a 40,000-page wiki behind `lilbee serve` with an empty models dir, so no engine started. A full lint reports 80,000 issues; cancelling one mid-scan left the server logging `Wiki lint cancelled after 14507 issues`.

Note for follow-up: `opencode run` interrupted with SIGINT exits without sending `notifications/cancelled`, so that path relies on the connection dropping rather than a protocol cancel.


tobocop2:
Correction to the note in this PR's description.

I wrote that `opencode run` interrupted with SIGINT exits without sending `notifications/cancelled`, implying a gap. That was measured badly: I inferred it from the absence of a cancellation line in the server log, but in that run the model had not yet dispatched the tool, so nothing was in flight to cancel.

Tested properly since: a client hard-killed with SIGKILL, sending no protocol cancel and only dropping the socket, still stops the work. A 40,000-page `wiki_lint` halted at 8,741 of 80,000 issues on the disconnect alone, against 14,507 for an explicit protocol cancel.

So both paths stop the work and there is no follow-up here. Whether opencode sends a cancel frame does not matter: the disconnect is enough.

PR #693, opened 2026-08-08
Collapse the demo reels in the README

tobocop2:
## What this does

Wraps the demo reels in `<details>` so the README skims. They stay gifs.

## What it does not do

An earlier commit on this branch converted them to `<video preload="none">` to defer loading. That does not work, and the claim was wrong: `raw.githubusercontent.com` serves mp4 as `application/octet-stream` with `x-content-type-options: nosniff`, so the browser is told the file is an opaque binary and forbidden from sniffing the real type. The player refuses it. Posters were fine because those come back as `image/png`.

Pages is not an alternative host: it publishes `site/` from `main`, so `lilbee.sh/demos/*.mp4` is a 404, and moving 73 MB of video under `site/` to fix that is worse than the problem.

Collapsing alone saves nothing either. GitHub adds no `loading="lazy"` to README images -- none of the 48 on the rendered page carry it -- and browsers fetch images inside a closed `<details>` anyway, because the content is in the DOM whether or not it is painted. So this is a readability change, not a bandwidth one.

## Found on the way

`site/index.html` and `site/tutorial/index.html` embed the same `raw.githubusercontent.com` mp4 URLs, so those players are broken for the same reason. Filed separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,7 @@ Point lilbee at a folder of PDFs, notes, ebooks, or code and it builds a searcha
 <details>
 <summary>Watch: /add a PDF, watch the Task Center, ask a cited question</summary>
 
-<video src="https://raw.githubusercontent.com/tobocop2/lilbee/gh-pages/demos/tui-add.mp4" poster="https://raw.githubusercontent.com/tobocop2/lilbee/gh-pages/demos/tui-add.png" controls muted playsinline preload="none" width="100%"></video>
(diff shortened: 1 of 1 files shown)


tobocop2:
Closing: this only collapses the gifs, and collapsing buys nothing. GitHub adds no loading=lazy to README images and browsers fetch images inside a closed details element anyway, so the page still pulls the same 72.7 MB. The deferral that would have justified it is not available: raw.githubusercontent.com serves mp4 as application/octet-stream with nosniff, so video players refuse it.

The real finding from this is filed as bb-dcc3o -- the marketing site embeds those same mp4 URLs, so its players are broken for the same reason.

PR #720, opened 2026-08-14
Agent-serving fixes: reasoning presentation and engine shape visibility

tobocop2:
## Problem

Two bugs make a served model look broken to an agentic client while the server behaves as designed. A reasoning model streams every token into `reasoning_content`, which clients like opencode never render, so the pane sits frozen for the whole think. And when the planner grants a smaller window or fewer slots than the configured target, nothing on the serve path says so, and the slot count is not exposed anywhere.

## Reasoning presentation

The translate layer takes a reasoning mode: `separate` (default, unchanged split), `inline` (thinking streams as ordinary `content` text on both paths, with the `<think>` markers stripped -- see the stacked #729), and `off` (asks the model not to think via `chat_template_kwargs.enable_thinking`; presentation falls back to the split for templates that ignore it). The mode comes from the new `completions_reasoning` setting, overridden per request by a `reasoning` field on the completions body. Non-string `reasoning` values (other vendors use the key for an effort object) are ignored and logged at debug, so no existing client request starts failing.

## Engine shape visibility

When the granted chat shape ends below the requested one, the fleet warns at engine adoption — the point every serving process passes through — naming granted vs requested slots and window. The granted shape is machine-readable: `served_chat_slots()` on providers, `chat_slots` on `/api/health`, and `slots` next to `context_window` on the active `/v1/models` entry. No sizing decision changes.

## Not covered

`lilbee status` stays engine-blind: it runs without a server and never warms the fleet; `/api/health` is the running-server surface.


tobocop2:
Validated end-to-end on hardware (trial merge of #720+#724+#721, 1x A100-80 + 2x A100-80 + 3x RTX 4090): all five reasoning-matrix cases pass including inline on the streaming path; the downsize warning fires for both the slot case (27B: 3 of 4 granted) and the window case (Qwen3-8B: 40960 of 65536); chat_slots lives on /api/health and /v1/models. The inline mode also produced the first recordable reel of a thinking model (Gemma 4 26B-A4B streaming its reasoning as visible text).

PR #721, opened 2026-08-14
Stop mis-sizing the KV cache for hybrid, per-layer-KV, and sliding-window models

tobocop2:
## Problem

The VRAM estimator mis-sizes the KV cache for several model families: hybrid-attention models are charged growing KV on every layer (Qwen3.6-27B Q4_K_M at one 65536 slot: 24.83 GiB estimated vs 17.6 GiB the engine allocates), models with per-layer KV-head counts are read from a scalar (LFM2 estimated zero KV), and sliding-window layers are charged full windows. The planner grants fewer slots and smaller windows than the card holds.

## Fix

Pin gguf-parser to fork tag `v0.25.0-lilbee1`: upstream main plus the three fixes proposed upstream as gpustack/gguf-parser-go#29 (hybrid `full_attention_interval`), #33 (per-layer `head_count_kv`), and #34 (SWA period table). Validated on an A40 against llama-server's own allocation logs: LFM2 and Qwen3.6-27B byte-exact, gemma3n / gpt-oss / gemma3 on the safe side, full-attention control unchanged.

## Revert path

The clone URL comes from `ENGINE_GGUF_PARSER_REPO` in `engine-versions.env`, so both build sites return to upstream with a one-line change once the PRs merge. The checked-in engine-wheel pin moves with the ref, which busts the engine build cache by design.


tobocop2:
Validated on hardware: estimator drops 24.83 to 18.60 GiB on Qwen3.6-27B Q4_K_M at one 65536 q8 slot (measured engine allocation 17.6 GiB), full-attention models unaffected, and the integration effect is real -- same model on the same A100, the old parser grants 3 slots, the fork grants 4x65536.

PR #724, opened 2026-08-14
Show first-turn prefill progress instead of nothing

tobocop2:
## Problem

An agent's first turn sends ~32k tokens of prompt, and on a large model that is 30 seconds to minutes of prompt processing after the launcher's warm bar ends. Nothing surfaces it: the client shows a spinner and no lilbee surface distinguishes working from hung (measured: 49 seconds of blank pane on Devstral 2 123B).

## Fix

llama-server streams `prompt_progress` frames when a streaming request opts in via `return_progress`. The fleet client now opts in on every streamed chat; the chat role's client reports (processed, total) to the provider and the same stream clears it when generation starts or the stream ends. `/api/health` carries `chat_prefill_processed` / `chat_prefill_total`, so polling health shows the prefill climbing.

## Docs

The agent-models guide states the first-prompt prefill cost plainly and points at the health fields. The warm-wait budget already scales with model weights; its existing test covers that.

## Not covered

Prompt-cache priming of the client's static prefix (an experiment, per the issue) and any in-client rendering; opencode has no channel for prefill progress.


tobocop2:
Validated end-to-end: /api/health carried a live prefill climbing 0 to 31216/31220 across 101 polls and cleared at first token. The fields also proved themselves in anger the same day, diagnosing a recording failure as an in-flight 24k cold prefill (2830/24099) that three other instruments missed.

PR #752, opened 2026-08-20
Warn about an overrun while the card can still hold it

tobocop2:
## Problem

The readback compares the engine's real footprint against the estimate and warns when they diverge. Its threshold was a flat 25%, while the margin that threshold protects is a setting.

Placement packs a card up to `cfg.usable_vram_fraction` and a single-card chat sizes its cache against `cfg.gpu_memory_fraction`. A load filling the binding share overflows once it exceeds its estimate by the remainder:

| usable_vram_fraction | overrun the card absorbs | fixed 25% threshold |
|---|---|---|
| 0.9 (stock) | 11% | overflows first |
| 0.95 | 5% | overflows first |
| 1.0 | 0% | overflows first |

So this is not only a problem once someone raises the setting. At stock values the warning already arrived after the overflow it exists to predict, on any card packed to its planning ceiling. Raising the fraction makes it worse, and that is done by exactly the people with the least room: its own docstring offers it as the way to serve a model that would otherwise be refused.

## Change

The overrun side takes the smaller of the old constant and three quarters of the room the binding fraction leaves. The binding fraction is `max(usable_vram_fraction, gpu_memory_fraction)`, since either can be raised past the other. At stock settings the threshold becomes 8.3%, inside the 11% the card can absorb.

## What keeps the flat tolerance

The shortfall side. An over-estimate costs capacity rather than the load, so narrowing it would add noise without preventing anything.

## Both paths

`report_divergence` and `_report_per_device` both take the bound. The per-device path reports the same overrun for a tensor-split load, and fixing only the first would have left the defect wherever a chat model spans cards.

## Verified on hardware

The threshold tightened from 25% to 8.3%, so the risk worth measuring is whether a correctly sized load now warns. On a RunPod A40 serving Qwen3-0.6B through `lilbee serve`, a real generation (`chat_ctx=16384`, HTTP 200, 4783 MiB resident) produced **zero** divergence warnings. Threshold tracking against the live config on the same machine: 0.9 -> 8.33%, 0.95 -> 3.95%, 1.0 -> 0.00%, and `gpu_memory_fraction=0.95` -> 3.95%, each at or below what the card absorbs. A measured +10% overrun warns where the flat 25% stayed silent.

Not measured: the live divergence ratio itself, only that it sat under 8.3%. This engine serves `GET /memory` rather than writing an engine log, so the log-parsing path was not the one exercised.

## Suite

`readback.py` at 100% coverage, 2063 tests green across the fleet/planning/placement/swap sweep, lint and mypy clean. A parametrized test asserts the threshold stays at or below the absorbable overrun across 20 combinations of the two settings, and restoring the fixed threshold fails the raised-fraction case.

### DIFF
--- a/src/lilbee/providers/fleet/readback.py
+++ b/src/lilbee/providers/fleet/readback.py
@@ -171,7 +171,8 @@ def report_divergence(
     if estimated_bytes <= 0 or actual_bytes <= 0:
         return False
     ratio = actual_bytes / estimated_bytes
-    if abs(ratio - 1.0) <= tolerance:
+    limit = min(tolerance, _absorbable_overrun()) if ratio > 1.0 else tolerance
+    if abs(ratio - 1.0) <= limit:
         return False
     log.warning(
         "The %s model %s allocated %.1f GiB of GPU memory but was planned for %.1f GiB "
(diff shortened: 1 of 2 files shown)


tobocop2:
CI is green: 18/18.

The earlier `integration (ubuntu-latest, 3.11)` failure was a setup flake, not this change. It died inside `./.github/actions/ollama-pull` before the test suite ran, with no Python error and no `readback` in the log, while `3.12` and `3.13` passed the identical step on the identical commit. It passed on rerun with no code change.

Hardware check: A40 serving Qwen3-0.6B through `lilbee serve`, real generation (`chat_ctx=16384`, HTTP 200, 4783 MiB resident), zero divergence warnings at the new 8.33% threshold. Threshold tracking on the same box: usable 0.9 -> 8.33%, 0.95 -> 3.95%, 1.0 -> 0.00%, and `gpu_memory_fraction=0.95` -> 3.95%, each at or below what the card absorbs.

One open question for the reviewer. This tightens the stock threshold from 25% to 8.33%. That follows from the arithmetic, and one real load did not trip it, but it is a single model on a single card: if the estimator is routinely more than 8% low on some model class, this gets chatty. The conservative alternative is to keep the flat 25% as a floor, so the threshold only tightens once `usable_vram_fraction` goes past 0.8. That still closes the config-dependent hole and leaves stock behaviour untouched.

PR #758, opened 2026-08-21
Refuse a model the engine cannot load before downloading it

tobocop2:
## Problem

lilbee judged a pull on the architecture string alone. Nothing looked at how the weights were quantized, so a file carrying a tensor type the bundled engine has no kernel for downloaded in full and failed at load. `prism-ml/Ternary-Bonsai-27B-gguf` reports architecture `qwen35`, which is supported, and packs its tensors as ggml type 42, which upstream does not define at all.

Separately, `enforce_arch_compat` reads the ref the user typed. A bare `owner/name` repo names no file, so that guard decided nothing and every unsupported architecture behind a bare ref downloaded in full.

## Solution

The check asks gguf-parser, which already ships beside llama-server and is built from the same pin, so the types it accept are the binary's rather than a table lilbee would have to keep in step. It reads the header over range requests and downloads none of the weights.

## Why the engine's own reader and not a generated list

`Luncy1978/Bonsai-8B-gguf` resolves to a `Q1_0` file. Upstream ggml does define that type id, so a list generated from upstream would have accepted it. The parser refuses it, because this build has no kernel for it.

## What counts as a verdict

A failure naming a tensor is the engine's answer. A 404, DNS failure or expired token is not, and leaves the download to report the real problem. An install without the engine extra has no parser to ask and pulls as before.

## Architecture, re-checked against the resolved file

The architecture is now judged on the file resolution actually chose rather than on the ref, which closes the bare-repo hole. Both verdicts sit behind `--allow-unsupported`, which skips them together.

## Evidence

Across the 240 most-downloaded GGUF repos over all four roles, running the shipped path: 226 accepted, 13 refused, 1 gated, no projector resolved as a model, and every accepted file's first bytes are `GGUF`. Before the architecture re-check, 9 of those 240 resolved to weights the engine cannot load.

Eleven Bonsai repos: 8 refused with the type named, 3 accepted, which are the community re-quantizations that really do load.

## Not covered

Refusal carries a message, not a machine-readable code. Unlike an unsupported architecture there is no preflight route and no list of accepted types to offer. `UnsupportedQuantError` subclasses `RuntimeError` so every pull surface renders it the way it already renders a failed pull, matching `ModelNotFoundError`.


tobocop2:
Back to draft. This must not merge as it stands: it would refuse a model that works.

The gate asks the bundled gguf-parser whether the engine can decode a file. That parser is pinned separately from llama-server and its type table stops at MXFP4, so it refuses `Q2_0` outright. `llama-server` from the pinned build loads `Ternary-Bonsai-1.7B-Q2_0_g64.gguf` and reports `model loaded`, and this branch would refuse to download it.

Two things are needed, and they are not independent.

**The parser has to know the types.** Fixed upstream in gpustack/gguf-parser-go#39, which adds NVFP4 (40), Q1_0 (41) and Q2_0 (42), corrects the alignment padding, and refuses a file whose declared tensors do not fit after the header. Until that lands, `ENGINE_GGUF_PARSER_REF` would have to point at the branch.

**The verdict has to drive selection, not gate the file already chosen.** Ranking orders Ternary Bonsai 27B as PQ2_0, Q2_0, Q2_g64. The check currently runs once on whatever resolution picked, so the pull resolves to PQ2_0, gets refused, and fails on a repo that contains a file lilbee can serve. Walking the ranking instead lands correctly:

| candidate | parser with the fix | outcome |
|---|---|---|
| `PQ2_0` | refused, type 142 unknown | skip |
| `Q2_0` (g128) | refused, tensor data does not fit | skip |
| `Q2_g64` | parses | install, and it runs |

That third row is also why the fit check is not optional here. Without it the g128 pack is accepted and about 7 GB of unloadable weights get downloaded.

Nothing in #757 needs reverting. Before it the picker installed a 600 MB projector as the model; after it the pick is real weights, just not the loadable variant. That is incompleteness, not regression, and its own fixes still hold.

tobocop2:
Both blockers are now addressed on the branch.

**The pin.** `ENGINE_GGUF_PARSER_REF` points at `v0.26.0-lilbee1`, that release plus the three ggml types llama.cpp defines above MXFP4, the alignment padding fix they depend on, and a check that a file's declared tensors fit after the header. Filed upstream as gpustack/gguf-parser-go#38 with PR #39; the pin returns to an upstream tag when that merges.

**The verdict now picks the file.** It ran once, on whatever resolution had already chosen, which lands on the wrong file whenever a repo publishes the same weights in several packings. `resolve_filename` takes the check as a parameter and walks its ranking until a candidate answers.

Measured against the real repos:

| repo | before | now | `llama-server` on the new pick |
|---|---|---|---|
| `prism-ml/Ternary-Bonsai-1.7B-gguf` | `PQ2_0` | `Q2_0_g64` | model loaded |
| `prism-ml/Ternary-Bonsai-27B-gguf` | `PQ2_0` | `Q2_g64` | same packing |
| `Qwen/Qwen3-0.6B-GGUF` | `Q8_0` | `Q8_0` | unchanged |
| `prism-ml/Bonsai-27B-gguf` | `Q1_0` | `Q1_0` | unchanged |

Writing it turned up a bug of the same shape as the one this PR exists to fix. A refusal was recognised by the phrase `read tensor info`, which matches only the unsupported-type error; the parser refuses a file whose tensors do not fit with different wording, so a real verdict read as no verdict and the file was accepted. A non-zero exit is now the verdict unless the message shows the parser never reached the file.

Still a draft: the upstream PR is open, so the pin is on a fork.

PR #760, opened 2026-08-23
Trade GPU layers for the context a chat model needs to answer

tobocop2:
## Problem

A chat model whose fitted window cannot hold a grounded prompt is refused. That is right for a model too big for the box, and wrong for one that would fit with a layer or two in system memory.

DeepSeek-R1-Distill-Qwen-1.5B Q4_K_M is 1.04 GiB and is refused on a 2 GiB card at 512 tokens. The same file serves a 16384-token window with all 28 of its blocks on the card and only the output layer off it. For a 151k-vocab model that one tensor is the whole difference.

Reported on a 2014 laptop, on a model that worked before 0.6.90b420. The window sizing moved to gguf-parser in that window, whose footprint adds a compute buffer the older header math did not charge. The newer estimate is the more accurate one; what was missing is the fallback.

## The window and the offload are chosen together

`fit_chat_ctx` asked what window fit at the configured offload. It now asks which offload and window serve together and returns the pair as a `ChatFit`, so the launch runs the offload its window was sized against rather than recomputing one. The search bisects the layer grid: the estimate charges less the fewer layers the card holds, so the fitted window is monotone in the offload.

## Layers are traded only while the weights fit

Past that the trade serves the model from system memory, which is the unusable load the refusal exists to prevent. A 70B Q4_K_M wanting 40.5 GiB against a 36 GiB budget still refuses.

A `num_ctx` pin, a window the user capped below the minimum, a tensor split, and an architecture that reports no layer count all keep the configured offload.

## Validation

3080 cells: 154 text models x 10 card sizes x both memory models, against 990 before. 23 rescued, 0 lost, 0 unified cells changed. `served_config_moved` is 0, so in every cell where the base already served, the branch chose the identical offload and window and no working model has its offload quietly reduced.

Monotonicity, which the binary search depends on, was walked in full rather than sampled: 10 models x 4 cards, 1320 offload points. Adding a layer to the card never increased the fitted window, and replaying the search over each measured curve returns the true best offload on all 40 curves.

Of the 23 traded configurations, 20 of the 21 measurable were loaded into a live llama-server at exactly their chosen offload and window, with model, KV and compute measured on the device, and verified to allocate under budget at a tightest genuine margin of +198.8 MiB. Two resist measurement for different reasons: `bge-m3` is an embedding model the server refuses to load as chat, so it never reaches allocation, and `PLM-1.8B` trips a `GGML_ASSERT` inside the engine before it finishes loading. The one wrong rescue was `gemma-4-26B-A4B-eagle3`, which the second commit fixes.

## Not covered

Answering through a discrete GPU. The traded configurations were verified to allocate under budget on a live llama-server, not to serve a chat turn. The budgets are also modelled rather than physical: every load ran on one MI300X, so a 1 GiB or 6 GiB card was a number given to the planner, not hardware the allocator faced. The tightest margin, +198.8 MiB on Ling-3.0-tiny against a 4608 MiB budget, is exactly the case a real 6 GiB card would test, with its own fragmentation and driver reserve.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -597,7 +597,13 @@ flowchart LR
   overflow device 0 even when the cards are unequal. A chat on one card searches the
   same grid against one device (`ctx.fit_single_ctx`, entered through
   `engine_params.resolve_chat_ctx`), so both paths price the cache the model's
-  architecture holds. Header math (`model_cache.compute_dynamic_ctx`) is the fallback
+  architecture holds. On one card the search covers the **offload as well as the window**
+(`planning.fit_chat_ctx`): when full offload leaves KV room for less than a grounded
+prompt, layers move to system memory until the window is usable and the launch runs at
+that reduced `--n-gpu-layers`, so a small model on a small card is served instead of
+refused. The trade needs the weights alone to fit the budget; past that the model would
+answer out of host RAM, which is the unusable load `plan_launches` refuses on purpose.
+Header math (`model_cache.compute_dynamic_ctx`) is the fallback
   for when gguf-parser cannot answer; it charges every layer as dense attention over
   the whole window, which under-grants linear-attention, sliding-window and MLA models. On a
   single CPU/Metal box this
--- a/src/lilbee/providers/engine_params.py
+++ b/src/lilbee/providers/engine_params.py
@@ -9,6 +9,7 @@ provider compute them identically. No native binding.
 from __future__ import annotations
 
 import logging
+from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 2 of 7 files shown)


tobocop2:
Held out of 0.6.90b428.

The trade moves layers off the card, and gpustack/gguf-parser-go#45 reports that the compute buffer at partial offload is under-estimated by an amount that grows with model size: roughly 6 MiB on Qwen3-0.6B, 28 MiB on granite-3.0-3b-a800m, and 103 MiB on OLMoE-1B-7B. Partial offload is the region this change trades into, and the tightest verified margin here is 198.8 MiB. The two are the same order of magnitude, and only the error grows.

This merges once https://github.com/gpustack/gguf-parser-go/issues/45 is fixed.

PR #774, opened 2026-09-02
Validate the rewritten search query and answer document listing questions

tobocop2:
## Problem

Once a chat has history, the server rewrites the follow-up into a standalone search query with an extra chat-model call. It then used the first line of the reply as the query, with no validation. A small local model answers that prompt with a lead-in, a refusal, or its reasoning, so retrieval ran on text the user never typed. Nothing told the caller what was searched, so a wrong answer looked unexplainable.

Nothing enumerated the corpus either. "What files can you see?" took the ordinary retrieval path, so the answer named about seven documents whether the index held twenty or two thousand.

## Solution

`choose_retrieval_query` accepts a rewrite only when it shares a content word with the turn, holds one sentence, and stays near the length of the question. Any reject falls back to the question as typed, which costs the pronoun resolution and nothing else. The checks read punctuation and the turn's own vocabulary rather than a list of known lead-ins, so they hold for any language whose pack supplies stopwords.

The effective query now reaches the caller: `retrieval_query` on `/api/ask` and `/api/chat`, and a `retrieval_query` SSE event ahead of the answer.

`AggregateKind.SOURCE_LISTING` answers listing questions from the store and reports how many it listed against the true total. Topical questions such as "which documents mention the harbor?" still take the retrieval path.

## Not covered

A rewrite can still fold an unrelated conversation topic into a question that already stood alone. Removing that needs a pre-check that skips condensation when the question holds no unresolved referent.


tobocop2:
Closing. This was written before the user's logs arrived, and against them only one part holds up: surfacing the effective retrieval query. The listing route duplicates what lilbee status and the plugin's Browse documents already do, and the condense validator guards a reply shape no real model produced in testing (small models returned clean rewrites or nothing at all, and an empty reply already fell back to the raw question). The retrieval_query field and SSE event will land on their own in a small follow-up PR.

PR #775, opened 2026-09-03
Give each build of a release its own onefile cache directory

### LINKED ISSUE #237: Managed server crash
### Description

server exited (exit code 2)

### Steps to Reproduce

1. Install lilbee and plugin.
2. Configure models.
3. Observe all signals green.
4. Restart Obsidian later.

### Expected Behavior

No errors.

### Actual Behavior

Error message popup similar to what I posted from the log file. 

### Server Mode

### OS

Win 11

### Obsidian Version

1.13.7

### lilbee Plugin Version

0.6.91

### lilbee Server Version

0.6.90

tobocop2:
## Problem

Every build of a release extracts into the same onefile cache directory, `{CACHE_DIR}/lilbee/{VERSION}`. The bootstrap rewrites every cached file whose checksum differs from the payload. The default build and the CUDA build of one release have different payloads. When a machine holds both builds, each launch of the other build rewrites the files that differ.

On Windows that rewrite fails while the other build runs, because Windows locks a mapped DLL against writes:

```
Error, failed to open 'C:\Users\...\AppData\Local\lilbee\0.6.90.432\msvcp140.dll' for writing.
server exited (exit code 2)
```

One user hit this eight times in two days and never got a working server. The plugin installs the CUDA build on NVIDIA hosts. The scoop manifest installs the default build. One machine with both builds is a normal setup.

I reproduced the failure on a clean Windows Server 2025 VM with the v0.6.90b432 assets. The same build launched twice never fails. With the default build serving, a launch of the CUDA build fails with exactly that line.

## Solution

Key the cache directory by build, not by version alone. `ASSET_NAME` is unique per build, and every workflow already passes it. The spec becomes `{CACHE_DIR}/lilbee/{VERSION}-<asset>`. For example, `0.6.90.432-windows-x86_64-cu125` sits beside `0.6.90.432-windows-x86_64`. Each build keeps its cached fast start and never touches another build's files.

**Old caches go away.** At startup the compiled binary removes the extraction directories of other releases beside its own. It deletes only a sibling that holds the bootstrap manifest and carries a different version, so `bin/`, `models/` and the other build of the same release are never touched. A directory it cannot remove is logged and left in place.

Closes tobocop2/obsidian-lilbee#237.

### DIFF
--- a/src/lilbee/cli/app.py
+++ b/src/lilbee/cli/app.py
@@ -14,6 +14,7 @@ from lilbee.app.version import get_version
 from lilbee.cli.helpers import json_output as json_out
 from lilbee.core.config import cfg, config_load_error
 from lilbee.core.settings import overlay_persisted_settings
+from lilbee.runtime.onefile_cache import cleanup_stale_onefile_caches
 
 app = typer.Typer(help="lilbee: Local RAG knowledge base", invoke_without_command=True)
 console = Console()
@@ -196,6 +197,8 @@ def _default(
     )
     # basicConfig is a no-op when handlers already exist, so always set level explicitly
     logging.getLogger().setLevel(level)
+
+    cleanup_stale_onefile_caches()
 
     # Swallow lancedb's shutdown-time thread noise: opt-in side effect, not
     # imposed on library consumers of lilbee.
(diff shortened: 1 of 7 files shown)


tobocop2:

PR #776, opened 2026-09-03
Tell clients which query retrieval ran on

tobocop2:
## Problem

When a chat has history, the server rewrites the follow-up into a standalone query. Retrieval then runs on the rewrite, not on the text the user typed. The rewrite reached nobody: one DEBUG log line and nothing on the wire. A user whose answer went off-topic had no way to see that retrieval searched for something else. One user reported exactly that. The context "self-populated", and it was "hard to tell what is in the context at any given moment".

## Solution

**The effective query travels with the answer.** `RagContext` and `AskResult` carry `retrieval_query`. The searcher sets the field only when the rewrite replaced the question, so no surface compares strings. `/api/chat` returns it as `retrieval_query`. `/api/ask` carries the same field for parity; it is always null there, because `ask` runs without history. The chat stream emits a `retrieval_query` event with `{"query": ...}` before the first token. The ask stream shares the helper and never fires it. One shared helper emits the event for both streams, so they cannot drift. The accepted rewrite logs at INFO beside the other routing decisions.

**The TUI shows it too.** `ask_stream` yields a `RetrievalNotice` before the first token when a rewrite ran. It is the stream's first non-token event. The TUI shows "Searching for: ..." in the answer bubble's status row while the model works. `lilbee ask` handles the notice on stderr and `lilbee --json ask` carries `retrieval_query`, for parity; both run without history, so neither fires today.


tobocop2:

PR #777, opened 2026-09-03
Skip archives and images at discovery and cap chunks per file

tobocop2:
## Problem

Discovery accepted all 119 formats the extractor can open. It did not judge whether a format carries text worth embedding. One user cloned this repository's docs into his vault. Two gzipped benchmark files became 28,449 chunks over 45 minutes, twice. Discovery indexed SVG logos as images. His index was mostly machine ranking output. That is why every answer looked like it came from seven random documents.

## Solution

**Discovery refuses archives from the MIME data.** Discovery excludes a format when the extractor's MIME type names an archive subtype. `application/zip`, `application/x-tar` and `application/gzip` go. Structured containers such as `application/epub+zip` stay. A short deny-list covers what MIME cannot express, currently `.svg`. The sync summary reports every refused file as skipped, with the reason, on every sync while the file stays in the vault. Discovery refuses nothing silently. A refused file that an earlier version indexed leaves the index on the next sync and counts as removed. `lilbee add`, the MCP add tool and the upload route refuse the same formats at the entry point, with the same reason.

**One file cannot explode the index.** The new `max_chunks_per_file` setting (default 3000, `0` for no limit) stops a file after chunking and before the store embeds it. With semantic chunking on, the chunker spends its boundary embeddings before the limit fires. The summary records the file as skipped with the chunk count. The help text names the case for a higher value: a long document at a small chunk size. Every server settings surface registers the setting: the settings map, the TUI, the environment variable and the docs. The plugin settings tab does not show it yet.


tobocop2:

PR #778, opened 2026-09-03
Pin the engine to the chunk budget and log every launch

tobocop2:
## Problem

A warm or peer-held engine kept a stale embed window. The planner sizes the embed window from `chunk_size`. That key was not load-affecting. So an engine planned under one chunk size matched the signature for another, and the server adopted it with its old `token_cap`. I measured this on an L4. A server under the default chunk size adopted an engine planned under `chunk_size=126`. It embedded with a cap of 504 instead of 2048 and silently truncated dense chunks.

A fresh launch also logged nothing that names the llama-server binary, its build, the backend, or the probed devices. One user ran Vulkan on an NVIDIA card all day. The only trace was a device name inside an unrelated warning.

## Solution

**The chunk budget is part of the engine pin.** `chunk_size` joins the load-affecting keys. A different budget computes a different signature, and the server never adopts it. An in-process change drops and relaunches the fleet against the new window, on the same path that `num_ctx` and `kv_cache_type` take. That relaunch is correct: the field already requires a reindex.

**The placement pin covers the primary card and the vision slots.** `main_gpu` joins the placement keys, so a process that pins another card never adopts an engine that holds the model elsewhere. Each vision launch records the slot ceiling it was fitted against. A peer adopts the engine only when that ceiling covers its own `vision_ocr_concurrency`, the same rule the chat window already follows, so a memory-capped fit is never rebuilt in a loop.

**Every launch says what it launched.** One INFO line per instance names the binary, the engine build id, the backend and the devices. Adoption logs the same line plus the owner pid.


tobocop2:

PR #779, opened 2026-09-03
Turn off thinking in auxiliary calls and tighten the known-item match

tobocop2:
## Problem

**Auxiliary model calls left thinking on.** Query expansion, HyDE, history condensation, intent classification, entity extraction and memory extraction each run on a small token cap. Most of them left thinking on. A reasoning model spends the whole cap inside `<think>`, and the parser then strips it. The call returns nothing at full cost. In one user's day, nine of eighteen chat completions were expansion calls that produced zero variants.

**A two-letter word routed the whole query to one document.** When no filename matched a reference token by token, a unique substring hit still routed. In "the file we discussed", the pronoun after "file" became a reference. It matched "We" in `DSO Web Hosting.md` and replaced the entire retrieval context with that note.

## Solution

**One `aux_options` helper** carries the cap and `think: False` to every auxiliary call. No call site can forget it. The native llama-server and Ollama paths honor `think`. LM Studio and the hosted paths drop it.

**The output cap reaches every provider.** `generation_options` emitted `max_tokens`, a name the option filter dropped, so the user's cap never left the config. It now emits `num_predict`, the one name lilbee uses inside, and the provider translators rename it on the way out.

**Ollama turns thinking off too.** For an `ollama/` model the SDK path translates `think: False` into the reasoning setting litellm forwards, so auxiliary calls stop burning their cap there as well. LM Studio has no such switch.

**The loose match is a whole-token run.** `contains_reference` matches the reference as consecutive whole tokens of the filename stem. So "we" cannot match "Web", and "12" cannot match "2012". A quoted title still names its hyphenated file.

### DIFF
--- a/src/lilbee/app/memory.py
+++ b/src/lilbee/app/memory.py
@@ -26,6 +26,7 @@ from lilbee.data.store import (
     human_recall_predicate,
     local_owner_predicate,
 )
+from lilbee.providers.base import aux_options
 
 
 def make_memory_row(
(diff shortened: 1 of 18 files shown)


tobocop2:

PR #780, opened 2026-09-03
Count held-out files separately and show them in status

tobocop2:
## Problem

A note that yields no text gets a skip marker once. Obsidian creates such a zero-byte file for every clicked unresolved wikilink. On every later sync, the summary counted the file as **Unchanged**. The reconciliation guard saw only that run's failed and skipped sets. So it reported the same files as a "possible silent drop" every time. A user with twelve such notes saw three wrong signals. The document count was twelve short. The summary said everything was unchanged. The data-loss warning cried wolf. The reasons sat in `skip_reasons.json`, which nothing read back.

## Solution

**Held out is its own state.** A marker-held file gets a `held` verdict. The summary counts it separately with its reason, never as unchanged.

**The guard subtracts the held-out files.** Reconciliation counts the files this run held out beside its failed and skipped sets. It fires only for a file that no mechanism explains.

**Skipped files reach the user.** `/api/status` carries the held-out files with reasons, capped, with the true total. `lilbee status` prints the same list, and the TUI status screen shows a "Held out of the index" section with the same rows. A TUI sync that held files out says so and names the retry command.

**The status response counts documents.** `document_count` joins `/api/status`, so a client no longer reads the whole `sources` list for one number. That list leaves the response in a later release.

**An upload under a new name moves the old file.** When exactly one indexed file holds the same bytes as an upload, the server moves it to the new path instead of writing a copy. The next sync repoints the source, so a note that changed its upload key is indexed once.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -125,7 +125,7 @@ flowchart TD
     H --> HC{"hash matches<br/>the stored row?"}
     HC -->|yes| BF["unchanged —<br/>backfill the fresh stat"]
     HC -->|no| SK{"hash matches a<br/>failed-file skip marker?"}
-    SK -->|yes| SKIP["held out — failed here last sync;<br/>counted apart from unchanged,<br/>reported with its reason"]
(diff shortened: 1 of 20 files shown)


tobocop2:

PR #781, opened 2026-09-03
Save path settings as strings and report a failed write as 503

### LINKED ISSUE #238: Couldn't move lilbee storage
### Description

server exited (exit code 2)

### Steps to Reproduce

1. Install lilbee and plugin.
2. Configure models.
3. Observe all signals green.
4. Restart Obsidian later.

### Expected Behavior

No errors.

### Actual Behavior

Error message popup:

"Could not move lilbee storage into your vault: Server responded 500: ... Internal server error"

### Server Mode

### OS

Win 11

### Obsidian Version

1.13.7

### lilbee Plugin Version

0.6.91

### lilbee Server Version

0.6.90

tobocop2:
## Problem

`PATCH /api/config` with `documents_dir` or `vault_base` answered a bare 500 on every platform, with nothing in `server.log`. The plugin sends exactly that request at first managed setup when "store content in vault" is on. So every managed user since b420 saw "Could not move lilbee storage into your vault: Server responded 500". They saw it again on each start (#238 on the plugin).

Cause: `_apply_with_rollback` queues the validated value for persistence. For these two fields that value is a `Path`, and `tomli_w` refuses it (`Object of type 'PosixPath' is not TOML serializable`). The emitter switch in 25182957c dropped the stringification the old writer did. The route mapped only validation errors. The app registered no handler for other exceptions, so it logged nothing.

## Solution

**`save()` writes paths as strings** at the single write boundary. A round-trip test covers both fields. A route test asserts 200.

**A write failure names its cause.** `OSError` from the config write maps to a 503. Its detail names `config.toml` and the OS error. When the file is locked, it tells the user to close the program that holds it. The MCP `settings_set` and `settings_reset` tools return the same text.

**No silent 500s.** A catch-all handler logs unhandled exceptions at ERROR with the method and path. It returns the same JSON error shape.

Closes tobocop2/obsidian-lilbee#238.


tobocop2:

PR #783, opened 2026-09-04
Fix the onefile cache, archive ingest, engine pin, held-out reporting, and config writes

### LINKED ISSUE #237: Managed server crash

### LINKED ISSUE #238: Couldn't move lilbee storage

tobocop2:
## Problem

One Windows user with an RTX 4080 Super reported that lilbee showed only part of his vault, answered off topic, filled the chat context on its own, and crashed the managed server. His diagnostics traced every symptom to a server defect. This branch carries the ten fixes as one change. Each fix was reviewed and gated on its own pull request first; the table below links them.

## Solution

**Each build of a release gets its own onefile cache directory** (#775). Two builds of one version rewrote each other's files, which fails on Windows while the other build runs. Startup also removes the caches of older releases.

**Clients learn which query retrieval ran on** (#776). The rewritten follow-up travels with the answer on the API, the SSE stream, the TUI and the CLI.

**The follow-up rewrite is off by default and gated when on** (#784). The rewrite is a second model call whose output varies between runs. When the user turns it on, it runs only for a question that refers to earlier turns: a follow-up opener, a pronoun or reference word, or fewer than three tokens. A standalone question is searched as typed.

**The files inside an archive are indexed as their own sources** (#785). A zip, tar, gzip or 7z member becomes a source named `archive.zip/report.pdf`, nested archives add a path segment, and the per-file chunk cap applies to each member. A changed archive rewrites its members and removes the ones that are gone.

**The Windows engine ships the MSVC runtime beside it.** On a fresh Windows Server 2025 image every build faulted inside the older msvcp140.dll in System32 before listing a device. The bundle carries msvcp140, vcruntime140 and vcruntime140_1, and the bundle check requires them. The self-check also verifies its model download against the certifi bundle, since that image has an empty root store.

**Discovery refuses vector drawings** (#777). A gzipped benchmark file no longer becomes 28,449 chunks under one name. A file that chunks past `max_chunks_per_file` is skipped and reported. Add, upload and sync all refuse the same formats.

**The engine pin covers the chunk budget, the primary card and the vision slots** (#778). A warm engine with a stale embed window is never adopted. Every launch logs its binary, build, backend and devices.

**Auxiliary model calls run with thinking off and a real output cap** (#779). The known-item match no longer routes a two-letter word to one document.

**Held-out files are counted apart from unchanged and shown everywhere** (#780). Status carries `document_count`, the held-out list and its total. An upload that carries an indexed file's bytes moves that file instead of indexing a copy.

**Path settings save as strings and a failed config write answers 503** (#781). The MCP settings tools return the same text.

## Merged pull requests

| PR | Fix |
|---|---|
| #775 | One onefile cache directory per build, and cleanup of older releases |
| #776 | The query retrieval ran on reaches every client |
| #784 | Follow-up rewrite off by default; when on, only for questions that refer to earlier turns |
| #785 | The files inside an archive are indexed as their own sources |
| #777 | Archives and images refused at discovery; a chunk cap per file |
| #778 | Engine pin covers the chunk budget, the primary card, and the vision slots |
| #779 | Auxiliary calls with thinking off and a real output cap; tighter known-item match |
| #780 | Held-out files counted apart and shown everywhere; `document_count`; upload move |
| #781 | Path settings saved as strings; a failed config write answers 503 |

Each PR is closed as superseded; its review history stays on the PR.

Closes tobocop2/obsidian-lilbee#237.
Closes tobocop2/obsidian-lilbee#238.


tobocop2:
## Validation

| Check | Method | Result |
|---|---|---|
| Follow-up rewrite: same question with history, retrieval query and SSE event before and after | Two servers on the same index, the shipped 0.6.90b432 and this branch, driven over `/api/chat` and `/api/chat/stream` | Before: no field, no event. After: `retrieval_query` in the response and a `retrieval_query` event before the first token |
| Rewrite gate: a standalone question skips the rewrite, a pronoun follow-up runs it | Same setup with `history_rewrite` on | Passed |
| Windows: the onefile cache directory survives a second build, config writes | GCP Windows Server 2025 VM driven by startup script | Passed |
| Windows: the shipped engine crashed in msvcp140.dll on a fresh image; the runtime bundling fixes it | GCP L4 VM: crash reproduced on every build, VC++ redistributable install made the engine list CUDA0 and self-check pass, then the released cu125 build passed with the bundled DLLs | Passed |
| Windows: self-check download fails TLS verification on a fresh image | Same VM, then the certifi fix | Passed after the fix |
| Linux: NVIDIA driver present without nvidia-smi is detected as sandboxed | GCP L4 Ubuntu VM | Passed |
| Full gate: lint, format, typecheck, tests with 100% coverage | `make check` on the merged branch | Passed |
| CI | GitHub matrix | All cells passed except the Ollama warm-up segfault that also fails on main |
| Adversarial and convention review | Blast-radius review, build-vs-buy, AGENTS.md audit per superseded PR | Findings fixed before consolidation |

PR #784, opened 2026-09-04
Turn the follow-up rewrite off by default and gate it on questions that refer to earlier turns

tobocop2:
## Problem

With chat history, every turn runs a second chat-model call that rewrites the question into a standalone retrieval query, and retrieval searches the rewrite. The rewrite is model output. Two calls seconds apart on the same input produced two different queries. A standalone question pays the call too and can come back reworded. The user never sees which query retrieval ran on, so a drifted rewrite reads as the context filling itself with unrelated sources.

## Default

`history_rewrite` is now off by default. Retrieval searches the question as typed unless the user turns the rewrite on.

## Gate

When the rewrite is on, it runs only for a question that refers to earlier turns. The check is a deterministic pattern in the language pack: a follow-up opener at the start ("and", "but", "what about"), a pronoun or reference word anywhere ("it", "they", "that one", "the same"), or fewer than three tokens. A question that stands alone is searched as typed and makes no extra call. A false positive costs one rewrite call, which is the old behavior on every turn.

## Docs

The settings reference and the architecture page describe the new default and the gate.

## Measured

Same index, same 8B model, history says the Crown Victoria has a 4.6L V8, follow-up "How much oil does it take?". With the rewrite off, three of the four retrieved chunks came from a crawled hash-oil page and one from the manual. With the rewrite on, all four came from the manual. Turning the rewrite on is the right call for pronoun-heavy conversations, and the gate keeps it off the turns that do not need it.


tobocop2:
Superseded by #783, which now carries this change and lists it in its table.

PR #832, opened 2026-09-12
List recalled memories as marked sources

tobocop2:
## Problem

Recalled memory facts enter the system prompt on every grounded turn but appear in no sources list. The reader cannot tell they shaped the answer.

## Solution

Each recalled fact joins the turn sources as a marked row. Memory rows carry `memory_id` with a `memory:<id>` source, and render as plain labels with no file link. The streamed sources event keeps memories alongside the cited subset.

Contract for consumers: `sources[]` entries may now carry `memory_id`, and `source` values may use the `memory:<id>` form with `content_type` `memory`. Document sources are unchanged.


tobocop2:
Superseded by PR 853 (same content rebased onto tip after merge conflicts; old branch untouched).

PR #843, opened 2026-09-12
Flag safety-stripped models from HF tags

tobocop2:
## Problem

The picks rail recommends one model per role with no content signal. Two safety-stripped re-uploads fit a 32 GB machine today, and only alphabetical luck keeps one from becoming the single chat model a first-time user sees.

## Solution

The catalog derives a safety flag from each repo's HuggingFace tags and carries it on every row. Picks and the Discover For You rail exclude flagged rows. The flag ships as `safety_stripped` on the catalog response and the MCP browse entry, so the plugin filters on data instead of names.

Exclusion is picks-only by design. Browse and search still list flagged rows: an explicit search states intent, while a recommendation does not.

The exclusion is optional. A persistent `include_stripped_picks` setting, off by default, restores flagged rows to Picks and the Discover rail when set. A setting, not a per-request flag: the Discover rail is in-process, so only a setting reaches every recommendation surface, and a content preference does not vary per call.

## Notes

Plugin contract: the toggle writes `include_stripped_picks` through PATCH /api/config (or `LILBEE_INCLUDE_STRIPPED_PICKS`, or the TUI settings screen) and refetches; picks sections then include flagged rows, each still carrying `safety_stripped`. The plugin-side toggle is a separate change.


tobocop2:
Superseded by PR 854 (same content rebased onto tip plus the include_uncensored rename; old branch untouched).

PR #850, opened 2026-09-12
Drop the Ollama warm-up steps from CI

tobocop2:
## Problem

The per-PR job warms two models before any test runs. That step lost its purpose: the serving daemon now keeps models loaded on its own, and the first test absorbs the short cold load inside its own timeout.

## Solution

Both warm-up steps go away, with their retry loops. The daemon setup, the model pulls, the version floor, and the crash diagnostics all stay. If the runner ever crashes again, the first real inference surfaces it and the uploaded bundle still captures the evidence.

## Notes

Live-daemon coverage continues in the per-PR integration tests until the open Ollama-lanes PR moves it to the weekly scheduled QA lane. That PR deletes these same lines, so the two branches need a rebase against each other when either lands.


tobocop2:
Subsumed by PR 840, which landed first and deleted the same warm-up steps wholesale: the reworded comment lines no longer exist at tip either (verified by grep). The ollama#17205 diagnosis stands as this PR's record.

PR #868, opened 2026-09-18
Raise the quant size table to published file rates

tobocop2:
## Problem

Twelve of sixteen bytes-per-weight table entries read under the real published file, worst case 29% under, which tells someone a model fits in their RAM when it does not.

## Solution

Each entry now takes the highest published rate a checked-in corpus records for that label, plus a small margin, rather than one repo's figure. The corpus keeps only repos of 7B parameters or more, because a fixed-size vocabulary head inflates the rate on smaller models. An entry moves only where three distinct repos back it, and a new test fails if any entry drops below the corpus.

Most labels move 0.03-0.14 GB at typical sizes; Q2_K moves most, about 0.5 GB on an 8B model. The model picker reads this estimate, so it picks slightly smaller defaults at the RAM margin.

Also corrects a comment and docstring overstating a real file's cost against its type.

## Notes

Models below the floor still read low: a 120M bitnet build publishes at twice its Q2_K entry. Closing that needs a rate that scales with model size, not a bigger margin.

Stacked on the low-bit quant label fix; both touch the same sizing table. Now also carries that branch's fix for an installed-wheel crash: the label reader and the size estimate read a small product-owned table of ggml's type sizes instead of importing the dev-only `gguf` package.


tobocop2:
Closing this approach. The repo already ships two ways to read the exact file size and uses neither: the bundled gguf-parser runs before every download and returns the byte-exact `fileSize`, and the catalog's range probe already receives the total in its `Content-Range` reply. `huggingface_hub` is a hard dependency and the download path already HEADs for the real size.

The measurements in this branch are sound — spot-checks reproduce the published rates to six decimals — but a frozen per-model rate table reconstructs a number the product is already given, and it rots the moment a repo re-quantizes.

Replacement lands on #863's branch instead: read the real size where a file is in play, keep a small type-rate estimate for the browse list only, rendered as approximate.

PR #869, opened 2026-09-18
Fail the smoke crawl leg on an unreachable host

tobocop2:
## Problem

The artifact smoke test gates releases only: release candidate, release, and verify-release (twice), no plain CI. A recent change makes it skip the crawl leg with a warning when the target host is unreachable, so a permanently unreachable target would keep passing forever with no one told.

## Solution

An unreachable crawl target now fails the leg instead of skipping it. A scan of the last 30 runs of each release-shaped workflow found zero smoke-step failures caused by crawl unreachability, so the skip was guarding a case that never happened while quietly weakening a release gate. A missing curl (exit 127) still fails with its own distinct message. `SKIP_CRAWL=1`, the deliberate skip for artifacts built without the crawler extra, is untouched.

## Notes

Stacked on #864, which introduced the skip this fixes.


tobocop2:
Closing: this work is subsumed rather than pending.

This branch tightened a curl probe that its base branch added. That base no longer adds the probe — the reachability check is gone entirely and `tools/qa/artifact_smoke.sh` is now byte-identical to main, because the probe could not strengthen the gate: main already hard-fails an unreachable host through the 'crawled page not searchable' assertion, so both of the probe's branches ended in the same failure while buying a hard curl dependency and a second 20s timeout on every release run.

Rebasing this branch onto the current base would re-add the probe, the curl stub and the rerun machinery that base just deleted, so it would revert the fix rather than extend it.

The one thing worth keeping — test names that describe what they actually exercise — is already in the rewritten smoke suite on the base branch.

